PR #6, opened 2025-11-19
Migrate ProgressBarHandler internal methods to extensions, add ControlGallery Page, tests

drasticactions:
https://github.com/user-attachments/assets/e4776f4f-ff4c-4d06-a90c-8deafd755338

Matching what @jsuarezruiz did for BoxView, this migrates ProgressBarHandler code into extension methods. I've also added a Control Gallery page and tests.

I also added dataType to the ControlGallery samples as well. At first it didn't seem to work, but once I cleaned bin/obj I wasn't able to reproduce it, including removing the dataType, cleaning, rebuilding, adding it back, and building it again. I need to keep an eye out for it and if I can reproduce it again, I'll make an issue. This PR ControlGallery should work fully if checked out and if it fails, then we need to look further and make an issue.


samples/ControlGallery/ControlGallery/Pages/ProgressBarPage.xaml
```diff
@@ -0,0 +1,109 @@
...
+                <Border.StrokeShape>
+                    <RoundRectangle CornerRadius="10"/>
+                </Border.StrokeShape>
+                <VerticalStackLayout Spacing="10">
+                    <Label Text="Default ProgressBar"
+                           FontAttributes="Bold"/>
+                    <ProgressBar x:Name="DefaultProgressBar"
+                                 Progress="{Binding Progress}"
+                                 HeightRequest="20"/>
+                    <Label Text="{Binding ProgressPercent, StringFormat='Progress: {0}%'}"
```
jsuarezruiz: Can limit it to have two decimals (for example)?

drasticactions: Updated with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/6/commits/164fa932c92e3e075c9f20d24efa1e786eb6349f

Avalonia.Controls.Maui.tests.slnf
jsuarezruiz, line 1: Why removing it?

drasticactions: readded with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/6/commits/9e943f73319fd5a82bcaf7c8d9e0e023dc40e60b

I removed the slnf because I saw that it was duplicated with the nupkg slnf, but in retrospec those could either diverge more or I could change the Nuke files to use that slnf. And also it makes it clear that the slnf is there for holding the test projects to easily run them locally to.

.github/workflows/main.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dorny/paths-filter@v3
         id: filter
+        continue-on-error: true
```
drasticactions, line 22: This is to make sure if files are added that don't map to the directories, it won't error out.

jsuarezruiz (approved):

PR #8, opened 2025-11-20
Fix sample themes to use OS theme, add VSCode tasks

drasticactions:
image

I had hardcoded the theme by mistake for the sandbox and control gallery app. This fixes it so it should now properly use the right theme on the OS level. Later, we can add specific tests and pages for checking switching the theme dynamically.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,10 +6,10 @@
             "type": "coreclr",
             "request": "launch",
             "preLaunchTask": "build_sandbox",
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,34 @@
+{
```
drasticactions, line 1: I also added VSCode launch files to launch the sandbox and control gallery apps directly. It should also work with the C#/MAUI Dev Kit and the Solution Explorer, but this makes it easier to switch between the two projects.

jsuarezruiz (approved):

PR #17, opened 2025-11-21
Make custom Avalonia controls internal to lock them to MAUI ABI

jsuarezruiz:
This PR makes all custom Avalonia controls internal to prevent them from being used directly in standard Avalonia projects. These controls should only be accessible through MAUI handlers.

But due to C# accessibility constraints (public classes cannot inherit from generic base classes with internal type arguments), all the handlers using a custom Avalonia control were also made internal.


jsuarezruiz:
We will apply a different approach.

PR #27, opened 2025-11-25
Fix ApplicationHandler so it actually gets invoked

drasticactions:
This, I _think_ fixes the ApplicationHandler so it can work correctly. If you call `OpenWindow` and `CloseWindow` from `Microsoft.Maui.Controls.Application.Current` they will be respected and managed by the virtual platform view, and created within the app.

For WASM (And other SingleWindowAppLifecycle platforms), calling on OpenWindow and CloseWindow will do nothing.

### DIFF
--- a/src/Avalonia.Controls.Maui/Handlers/ApplicationHandler.cs
+++ b/src/Avalonia.Controls.Maui/Handlers/ApplicationHandler.cs
@@ -76,6 +76,9 @@ public partial class ApplicationHandler : ElementHandler<IApplication, Applicati
         {
             var activationState = new ActivationState(mauiContext, request.State ?? new PersistedState());
             var window = application.CreateWindow(activationState);
+
+            // For WASM, this will return as MauiAvaloniaContent from SingleViewWindowHandler, so it will not show.
+            // We don't support multiple windows in WASM in Avalonia currently.
             var avaloniaWindow = window.ToPlatform(mauiContext) as global::Avalonia.Controls.Window;
             avaloniaWindow?.Show();
         }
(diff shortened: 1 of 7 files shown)


src/Avalonia.Controls.Maui/Handlers/ApplicationHandler.cs
```diff
@@ -71,8 +72,13 @@ public static void MapTerminate(ApplicationHandler handler, IApplication applica
...
     public static void MapOpenWindow(ApplicationHandler handler, IApplication application, object? args)
     {
-        // Window opening is handled by the WindowHandler
-        // This is called after the window is created
+        if (args is OpenWindowRequest request && handler.MauiContext is IMauiContext mauiContext)
+        {
+            var activationState = new ActivationState(mauiContext, request.State ?? new PersistedState());
+            var window = application.CreateWindow(activationState);
+            var avaloniaWindow = window.ToPlatform(mauiContext) as global::Avalonia.Controls.Window;
+            avaloniaWindow?.Show();
```
jsuarezruiz, line 83: Should this use a compilation directive or detect if running in WASM at runtime to avoid it?

jsuarezruiz: And if we added a condition here; can we include a comment indicating the status in WASM? To remember it for the future.

jsuarezruiz (approved):

src/Avalonia.Controls.Maui/Handlers/ApplicationHandler.cs
drasticactions: Added a comment with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/27/commits/d029391fbde0e8dc380e3d98e8d39adea5f0a1b4

IMO I rather not specifically use multiple TFMs yet until we really need to. SingleViewWindowHandler would be used for any platform that uses SingleView Lifetimes (Like WASM, Avalonia.iOS, Android I think too, etc), and those return content rather than windows. In the future, we could change the default Avalonia lifetimes for mobile platforms to support multiple windows (I can see that happening for iOS to properly support Scene Delegates), so I rather not bind it at a TFM level for what gets returned, but either handle the output or skip it at the platform level.

PR #34, opened 2025-11-27
Updated ShapeViewHandlers implementation, added samples and tests 

jsuarezruiz:
This PR refactor all the shape related handlers to use extension methods, apply fixes in some properties like Path Transforms.

Added samples:
shapesviewhandler

And added several tests:
image

### DIFF
--- a/src/Avalonia.Controls.Maui/Extensions/ShapeViewExtensions.cs
+++ b/src/Avalonia.Controls.Maui/Extensions/ShapeViewExtensions.cs
@@ -287,36 +287,57 @@ public static class ShapeViewExtensions
     /// <param name="path">The virtual path view.</param>
     public static void UpdateRenderTransform(this AvaloniaPath platformView, PathShape path)
     {
+        if (platformView is null)
+            return;
+
         if (path?.RenderTransform == null)
         {
             platformView.RenderTransform = null;
             return;
         }
 
-        var matrix = path.RenderTransform.Value;
+        platformView.RenderTransform = ConvertTransform(path.RenderTransform);
+    }
 
-        if (path.RenderTransform is RotateTransform rotate)
+    private static global::Avalonia.Media.Transform? ConvertTransform(Transform transform)
+    {
+        switch (transform)
         {
-            platformView.RenderTransform = new global::Avalonia.Media.RotateTransform
-            {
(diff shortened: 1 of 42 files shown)


src/Avalonia.Controls.Maui/Extensions/ShapeViewExtensions.cs
```diff
@@ -0,0 +1,358 @@
...
+    /// <summary>
+    /// Updates the shape. Base implementation does not require additional work.
+    /// </summary>
+    /// <param name="platformView">Avalonia shape instance.</param>
+    /// <param name="shapeView">The virtual shape view.</param>
+    public static void UpdateShape(this PlatformShape platformView, IShapeView shapeView)
+    {
+        _ = platformView;
+        _ = shapeView;
+    }
```
drasticactions, line 41: Are these intended to be no-opt?

src/Avalonia.Controls.Maui/Extensions/ShapeViewExtensions.cs
```diff
@@ -0,0 +1,358 @@
...
+    /// <summary>
+    /// Updates the stroke miter limit. Avalonia does not expose this directly.
+    /// </summary>
+    /// <param name="platformView">Avalonia shape instance.</param>
+    /// <param name="shapeView">The virtual shape view.</param>
+    [NotImplemented("Avalonia Shape doesn't expose StrokeMiterLimit property.")]
+    public static void UpdateStrokeMiterLimit(this PlatformShape platformView, IShapeView shapeView)
+    {
+        
+    }
```
drasticactions, line 164: Since you landed StrokeMiterLimit with https://github.com/AvaloniaUI/Avalonia/pull/20156, after it lands on the nightly feeds, you should be able to bump the Avalonia version in the Directory targets to add it.

drasticactions: Actually, scratch that. You can't yet because of the Avalonia v12 net10 bump. We need to target this specific version we have because it still supports net8, and we can't bump to net10 until the MAUI changes land in a public build that I can bump to.

So... yeah, oops, sorry, lol.

drasticactions (approved):

PR #38, opened 2025-11-28
Updated SwitchHandler implementation and added samples

jsuarezruiz:
This PR updates the SwitchHandler moving the mapping logic to extension methods, fixes the implementation assigning colors (Track and Thumb) and add samples.

switchhandler


drasticactions (approved):
LGTM 👍

PR #41, opened 2025-12-02
Updated PickerHandler implementation, added samples and tests

jsuarezruiz:
Updated the PickerHandler implementation moving mapping logic to extension methods, adding samples and tests.

image

NOTE: Draft PR, working on fixes some properties, for example, _SelectedIndex_ is not working yet.


jsuarezruiz:
We ended up rejecting https://github.com/AvaloniaUI/Avalonia/pull/20199, and instead implemented the needed functionality by introducing a new control derived from ComboBox that adds Header support. This aligns with what the PickerHandler expects.
Here’s the current implementation in action:
image

drasticactions (approved):
Since I merged the Warning-as-Error change I fixed the IL Warnings and set it to use a FunDataTemplate instead.

I had asked Max to review since I knew ya'll were talking about the construction of the control. Since there are tests here I'm going to pull it in as an initial implementation. We can iterate on the composition to make it better should we need to.

PR #44, opened 2025-12-03
Refactor ImageHandler, added tests and samples

jsuarezruiz:
Updated the ImageHandler implementation moving mapping logic to extension methods, adding samples and tests.

imagehandler

Implemented:
- `IsLoading` property
- `IsAnimationPlaying` property
- Gif support
- Image Cache

Because the Avalonia Image control not support directly gif, the handler manage internally two controls: Image and GifImage from Avalonia.Labs.

image


drasticactions (approved):
As an initial refactor this LGTM.

I wrote an issue for setting the UserAgent for HttpClient and not setting the default (https://github.com/AvaloniaUI/Avalonia.Controls.Maui/issues/45), and I need to look into if we can leverage IHttpClientFactory instead for creating HttpClient instances. What was originally there was needed to get _any_ images loaded for the sake of the POC.

What's here is better, with room to iterate later once we can figure out the specs of what it should do.

PR #49, opened 2025-12-04
Updated GraphicsViewHandler implementation, added samples and tests

jsuarezruiz:
Updated GraphicsViewHandler implementation to move mapping logic to public extension methods.

graphicsviewhandler

Added samples and tests.

image


drasticactions (approved):
LGTM 👍

PR #52, opened 2025-12-04
Set Window Width/Height to ClientSize if MAUI's is NaN, always do PlatformArrange on UIThread

drasticactions:
https://github.com/user-attachments/assets/d827f6ae-79ab-49e9-8528-6c31b1b16391

When launching the MAUI app, if you didn't resize the window it could cause Avalonia Layout Cycles to happen. Stepping through it, the layout calculations stem from the Avalonia Window Width/Height values, which were NaN, and then causing the inner layouts to do layout cycles and break.

Setting the initial Window Layout size to either be what the user set on the MAUI Window itself on launch, or the Avalonia ClientSize if that is null, means it should always have a proper size on launch. That fixes the control issues as far as I can tell.

Also, PlatformArrange should always be on the UI Thread as far as I can tell (I don't think there's ever a case where it shouldn't) and that was breaking stuff like the Word Puzzle, since the Animation Manager was sending events on background threads. That needs more research, but I _think_ protecting it here should be fine.


src/Avalonia.Controls.Maui/Handlers/ViewHandler.cs
```diff
@@ -164,8 +164,10 @@ public virtual void PlatformArrange(Microsoft.Maui.Graphics.Rect frame)
     {
         if (PlatformView is null)
             return;
-        PlatformView.Measure(new global::Avalonia.Size(frame.Width, frame.Height));
-        PlatformView.Arrange(new global::Avalonia.Rect(frame.X, frame.Y, frame.Width, frame.Height));
+        Avalonia.Threading.Dispatcher.UIThread.Invoke(new Action(() => {
```
maxkatz6: Does it require `Dispatcher.UIThread.CheckAccess` calls? Just to avoid DispatcherOperation flow for the happy path.

src/Avalonia.Controls.Maui/Handlers/WindowHandler.cs
```diff
@@ -96,13 +100,19 @@ static void mapY(IWindowHandler handler, IWindow window)
     static void mapWidth(IWindowHandler handler, IWindow window)
     {
         var avWindow = GetWindow(handler, window);
-        avWindow.Width = window.Width;
+        if (!double.IsNaN(window.Width))
+            avWindow.Width = window.Width;
+        else
+            avWindow.Width = avWindow.ClientSize.Width;
```
maxkatz6, line 106: I wonder if it should be `avWindow.ClearValue(Window.WidthProperty)` - this call will reset to whatever default it had.

drasticactions: image

The Width/Height values are still NaN after being called. Maybe it's an Avalonia bug?

src/Avalonia.Controls.Maui/Handlers/ViewHandler.cs
drasticactions: Added with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/52/commits/f43aac932cc7e743dd0557684ebb3c4e48acb18f

src/Avalonia.Controls.Maui/Handlers/WindowHandler.cs
maxkatz6: I see. I thought you can remove your overwritten width/height value this way, but original width/height values set from clientsize do not exist at this point. Not a bug really.

maxkatz6: Setting ClientSize seems to be what Window.cs does itself
https://github.com/AvaloniaUI/Avalonia/blob/fd37190b324dbe70abb55fa40342dc4ebda23017/src/Avalonia.Controls/Window.cs#L1207

drasticactions:
Max pointed out this code:

https://github.com/AvaloniaUI/Avalonia/blob/fd37190b324dbe70abb55fa40342dc4ebda23017/src/Avalonia.Controls/Window.cs#L1207-L1208
https://github.com/AvaloniaUI/Avalonia/blob/fd37190b324dbe70abb55fa40342dc4ebda23017/src/Avalonia.Controls/Window.cs#L828-L831

and there is also https://github.com/AvaloniaUI/Avalonia/issues/18060

Thinking it through, I think this needs more research, but for the sake of the mapper, I think making sure that the window should always be set to a valid value _should_ be okay, and using the ClientSize would map to what Avalonia does for windows. But it's odd that the value isn't already set when it appears it should be.

PR #55, opened 2025-12-05
Added pending BoxView sample in the Gallery

jsuarezruiz:
We already had the handler with tests and the status updated. This PR just include a sample plate in the Gallery.

boxview-sample


drasticactions (approved):
LGTM 👍

PR #58, opened 2025-12-05
Implement SliderThumbImageSource to Slider

jsuarezruiz:
After merge https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/50 is time to review some ImageSource properties in control and implement it.

Implement `SliderThumbImageSource` property to Slider.

slider-thumbimagesource

And tests:
image


jsuarezruiz:
Updates the Status markdown too.

src/Avalonia.Controls.Maui/Extensions/SliderExtensions.cs
```diff
@@ -282,4 +306,56 @@ void OnTemplateApplied(object? sender, TemplateAppliedEventArgs e)
...
+                        slider.GetVisualDescendants().OfType<Thumb>().FirstOrDefault();
+
+            if (thumb == null)
+                return;
+
+            var size = image.Size;
+            var width = size.Width > 0 && !double.IsNaN(size.Width) ? size.Width : ThumbSizeFallback;
+            var height = size.Height > 0 && !double.IsNaN(size.Height) ? size.Height : ThumbSizeFallback;
+
+            // Clamp the thumb size to 40–64 px to ensure consistent visuals across images:
```
jsuarezruiz: Question: Should we have a logic to auto-resize huge images? or require to use an image with the correct size directly?

drasticactions: I would try to match what MAUI itself does for the native platforms where possible, to be consistent with it. If it doesn't resize images, we shouldn't either. If it should, then the logic for what the size should be should _probably_ be in MAUI and we apply those sizes. 

I think that would also be an issue with Tab Bar controls as well with images. I think there was a bug in MAUI in the past for the same complaint that images were not being scaled down, so they should be huge.

jsuarezruiz: Applied changes to use the original image sizes and have exactly the same .NET MAUI behavior.

drasticactions:
I updated it to move the slider image to the new, standard, location. 

image

LGTM 👍

drasticactions (approved):

PR #60, opened 2025-12-05
ImageButtonHandler, AnimationManager, Fix Tests, CI...

### LINKED ISSUE #56: Turn on Warnings as Errors for core Libraries.
Basically to make sure we don't introduce trimmer errors and stuff like that. We might want to bring in the editor config and style cop to enforce formatting as well.

drasticactions:
This ended up being a bigger PR than I thought.

- Moves ImageButtonHandler/MauiImageButton into the new locations and refactors them.
- Adds new tests for them
- Adds an updates ControlGallery entry for them.

- Add MauiPlanets demo app. It should work except for the CollectionView images in the gallery not loading. That needs more work.
- But that led to finding out that we need to also implement the AnimationManager to get proper animations on the UI Thread for Avalonia. I did that and have tests for it...

- Which then led me to find out that tests were failing in CI and ended up green, because I am dumb and set it to Continue On Fail, thinking that posting the results would cause it to fail, which it doesn't. I need to let it still fail. The results will post after.

I then fixed the failing tests, which were either handles not being created for stubs, and the use of Reflection for the ButtonHandler tests for getting the content. We should never use reflection in the standard library, so I put skip on those tests until we can come up with a better solution.

And I turned on Treating Warnings as Failures for Avalonia.Controls.Maui and Avalonia.Controls.Maui.Tests, fixes #56.


samples/ControlGallery/ControlGallery/Pages/ButtonPage.xaml
```diff
@@ -476,7 +476,471 @@
                             TextColor="Black"/>
                 </VerticalStackLayout>
             </Border>
-            
+
+            <!-- ImageButton Section -->
+            <Label Text="ImageButton Control"
```
jsuarezruiz: Could we separate in a different Page and include a new `TextCell` in the MainPage?

samples/ControlGallery/ControlGallery/Pages/ButtonPage.xaml
```diff
@@ -476,7 +476,471 @@
...
+                            <Label Text="Simple Command"
+                                   FontSize="10"
+                                   HorizontalOptions="Center"/>
+                        </VerticalStackLayout>
+
+                        <VerticalStackLayout>
+                            <ImageButton Source="dotnet_bot.png"
+                                         WidthRequest="70"
+                                         HeightRequest="70"
+                                         Command="{Binding ImageButtonParameterizedCommand}"
```
jsuarezruiz: Cannot see updates in the Label? Can review it?

jsuarezruiz (changes requested):
Should we create a Samples (or similar name: Showcase, etc) folder to group projects with samples?

There is a clipping issue in the profile page:
Image
Will not pay a lot of attention to it because use a Border and is implemented but we need to review it and complete it (refactor handler, add tests etc).

Also the fake tabbar is on top of the planets lists (can fix easily in the sample).

Then, in details, the Gallery not display anything:

Image
Use a CollectionView and could happen something similar to the Border. 

Once we work around Border and CollectionView, this sample should work nicely.

samples/ControlGallery/ControlGallery/Pages/ButtonPage.xaml
drasticactions: Fixed with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/60/commits/911a1c9851ef4ec63fd19318fbfcb7d806e12cdf

samples/ControlGallery/ControlGallery/Pages/ButtonPage.xaml
drasticactions: Fixed with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/60/commits/513049aa6860830c824f67c0e74f25f5d4fb8490

drasticactions:
@jsuarezruiz I moved MauiPlanets to a `showcase` directory, and refactored it to removing the build warnings and update the APIs.The images should now be working.

スクリーンショット 2025-12-06 10 19 45

I also refactored ControlGallery and MauiPlanets to get all of the resources into the class library and reference them from there, cutting down on the duplicated files.

drasticactions:
@jsuarezruiz 
スクリーンショット 2025-12-06 11 10 47

https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/60/commits/f713af2bc60574fc06ed55aaaf530cb8b111920c

Applying the clip to Image should fix that.

PR #69, opened 2025-12-12
Updated RadioButtonHandler implementation, added samples and tests

jsuarezruiz:
Updated **RadioButtonHandler** implementation moving all the mapping logic to public extension methods. Completed everything and updated Status.

Added a specific sample page in the gallery:
radiobuttonhandler

And tests:
image


jsuarezruiz:
NOTE: Is a Draft because want to explore if can avoid the `MauiRadioButton` logic (to manage custom ControlTemplates correctly) and use directly the Avalonia `RadioButton`.

src/Avalonia.Controls.Maui/Extensions/AvaloniaResourcesBuilderExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+/// </summary>
+public static class AvaloniaResourcesBuilderExtensions
+{
+    private const string RegistrationKey = "__AvaloniaControlsMauiResourcesRegistered";
+
+    /// <summary>
+    /// Registers the built-in Avalonia themes and all control templates/styles
+    /// shipped with Avalonia.Controls.Maui.
+    /// </summary>
+    public static MauiAppBuilder UseAvaloniaResources(this MauiAppBuilder builder)
```
jsuarezruiz: `ActivityIndicator` (for example) is not rendering, are we not missing this part in Desktop?

drasticactions: I want to know for sure if this code is actually needed for ResourceDictionaries, or if we should write these themes another way.

If it's needed, it can't be in yet another builder to be added to MauiAppBuilder, it should be in UseAvalonia. But ideally, it shouldn't be needed at all, and we're doing this totally wrong.

drasticactions: @maxkatz6 Can you please review this? What do you think?

src/Avalonia.Controls.Maui/Controls/RadioButton/MauiRadioButton.axaml.cs
```diff
@@ -0,0 +1,11 @@
+using Avalonia.Markup.Xaml;
+
+namespace Avalonia.Controls.Maui;
+
+public partial class MauiRadioButtonResources : ResourceDictionary
```
maxkatz6, line 5: Instead of adding x:Class code behind for each dictionary, you can include these in `ControlStyles` XAML directly 

Pseudocode:
```xaml
<Styles>
    <StyleInclude Source="ExistingIncludes.axaml" />

    <Styles.Resources>
        <ResourceDictionary>
             <ResourceDictionary.MergedDictionaries>
                  <ResourceInclude Source="avares://PathTo/MauiRadioButton.axaml" />
```

maxkatz6: You essentially only need single x:Class for all styling, which will root all other styles/resources from being trimmed.

src/Avalonia.Controls.Maui/Styles/Controls.axaml.cs
```diff
@@ -0,0 +1,12 @@
+using Avalonia.Markup.Xaml;
+using Avalonia.Styling;
+
+namespace Avalonia.Controls.Maui;
+
+public partial class ControlStyles : Styles
+{
+    public ControlStyles()
+    {
+        AvaloniaXamlLoader.Load(this);
```
maxkatz6, line 10: Not strictly necessary here, since this file is inlcuded only from C# code.
But if it was used from XAML, it would be correct to pass `IServiceProvider` from ctor to AvaloniaXamlLoader.Load call:

https://github.com/AvaloniaUI/Avalonia/blob/f46758f972f71531319957f075b9d5d334343bbe/src/Avalonia.Themes.Fluent/FluentTheme.xaml.cs#L28-L30

This will add context awareness for nested markup extensions, if any.

src/Avalonia.Controls.Maui/Extensions/AvaloniaResourcesBuilderExtensions.cs
maxkatz6: Framework initialization is structured a bit weirdly here, from what I see. Correct me if I am wrong here:

1. Program.cs creates an Avalonia app, as it would usually be in normal desktop project
2. Avalonia app must inherit MauiAvaloniaApplication
3. Avalonia app _should_ include a theme for avalonia controls. For example - SimpleTheme.
4. MauiAvaloniaApplication then instanciates MAUI application
5. MAUI application does its setup and calls RegisterResources from OnLaunching
6. RegisterResources registers another set of Avalonia themes, now specific to controls unique to MAUI, not included in SimpleTheme

There is a logical link between step 3 and 6, yet these are far separated.

maxkatz6: Alternatives I see:

1. Define `MauiControlStyles` (ControlStyles from this PR), and include it near the `<SimpleTheme />` in each users App.axaml. Cons: versbose. Pros: flexibility?.
 
2. Redefine MauiAvaloniaApplication to automatically include MauiContorlStyles in a base class implementation. Possibly from OnFrameworkInitializationCompleted. Less verbose. A bit less flexibility

3. Rethink need of MauiAvaloniaApplication. It could be a `WithMauiApp<TMauiApp>(this AppBuilder appBuilder)` avalonia builder extension. Essentially moves majority of code from MauiAvaloniaApplication to this extension method, allowing user to define any kind of Avalonia applicaiton (not necessary implementing MauiAvaloniaApplication).

maxkatz6: I would probably prefer 1 or 2 for simplicity

maxkatz6: XPF does option 3. Not because it's better, just happened to be this way.

You can see this [UseAvaloniaXpf](https://github.com/AvaloniaUI/Xpf/blob/7db54d43f3cb1c12aec4db293f738f0b40f0b78c/src/Atlantis/AppBuilderExtensions.cs#L14-L24) extension.
It set's up basic necessities for `XPF`, and `XPF` specific themes to the application (with `Insert(index: 0)`, so it can be overriden by user defined XAML).

And then there is `DefaultXpfAvaloniaApplication`. Which is used by default, but is completely optional. [XAML](https://github.com/AvaloniaUI/Xpf/blob/7db54d43f3cb1c12aec4db293f738f0b40f0b78c/src/Atlantis/DefaultXpfAvaloniaApplication.axaml) part and [C#](https://github.com/AvaloniaUI/Xpf/blob/7db54d43f3cb1c12aec4db293f738f0b40f0b78c/src/Atlantis/DefaultXpfAvaloniaApplication.axaml.cs) part.

jsuarezruiz: Thanks @max, this breakdown is helpful.
I think the key principles we should optimize for are:
- Minimal ceremony for developers
- Clear ownership of where styles come from
- Predictable override order

Given that, I’d lean toward a balanced version of option 2. Having `MauiAvaloniaApplication` automatically include the MAUI control styles (`OnFrameworkInitializationCompleted`):
- A working app by default (ActivityIndicator rendering, etc.)
- No requirement for users to understand or manually wire theme layers
- A single, obvious place where “MAUI-on-Avalonia” behavior lives

Also, users can still override everything in their own `App.axaml`.

src/Avalonia.Controls.Maui/Platform/MauiAvaloniaApplication.cs
```diff
@@ -134,4 +136,20 @@ private void OnActualThemeVariantChanged(object? sender, EventArgs e)
...
     }
+
+    private void RegisterMauiControlsResources()
+    {
+        const string RegistrationKey = "__AvaloniaControlsMauiResourcesRegistered";
+
+        if (Resources.ContainsKey(RegistrationKey))
+            return;
+
+        Resources[RegistrationKey] = true;
```
drasticactions: I _think_ OnFrameworkInitializedCompleted only gets called once, so this key check wouldn't be needed, but I don't know enough about how that gets called to know for sure. @maxkatz6 What do you think?

maxkatz6: Yes, this key is not necessary. OnFrameworkInitializedCompleted  will be called once

drasticactions (approved):
Now we should be good to go!

PR #70, opened 2025-12-12
Create full Axaml/Axaml.cs controls for ComboBox/ProgressRing/RadioButton

drasticactions:
I had broke the themes for MauiComboBox/ProgressRing/MauiRadioButton when I was trying to fix the WASM / Trimmed builds, since the resource dictionary that were included would blow up since they didn't exist. @jsuarezruiz adds it back with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/69, but that we keep hitting this doesn't sit right with me.

I went and converted these controls to be complete Axaml files with code-behind, rather than Resource Dictionaries that happened to reference C# versions of the controls. This seems to fix it right in all mode: Release, debug, NativeAOT, WASM.


src/Avalonia.Controls.Maui/Controls/ProgressRing/ProgressRing.axaml
```diff
@@ -1,63 +1,70 @@
...
+    x:Class="Avalonia.Controls.Maui.ProgressRing">
 
-    <ResourceDictionary.ThemeDictionaries>
-        <ResourceDictionary x:Key="Light">
-            <SolidColorBrush x:Key="ProgressRingTrackBrush" Color="#20000000" />
+    <ProgressBar.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.ThemeDictionaries>
+                <ResourceDictionary x:Key="Light">
+                    <SolidColorBrush x:Key="ProgressRingTrackBrush" Color="#20000000" />
```
jsuarezruiz, line 11: These resources are now instance-scoped, not theme/global scoped. Every `MauiComboBox` instance will allocate its own brushes and thicknesses. In this particular case is correct, but is something can be shared, move to Styles dictionary.

jsuarezruiz (approved):

PR #71, opened 2025-12-15
Add MergedDictionaries for ControlTheme to default AppBuilder

drasticactions:
スクリーンショット 2025-12-15 16 56 13

Attempt number 2 from https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/70,

This splits out stuff done in https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/69, except

- It's apart of the default `UseAvalonia` path, so no need to add another builder parameter.
- No `Control.axaml` (https://github.com/AvaloniaUI/Avalonia.Controls.Maui/blob/1a7f180dce6fd6d8830c451e69f9db528f98833c/src/Avalonia.Controls.Maui/Styles/Controls.axaml) which, as far as I can tell, is redundant when directly adding the themes in code and would be trimmed out as written.

### DIFF
--- a/src/Avalonia.Controls.Maui/Extensions/MauiAppBuilderExtensions.cs
+++ b/src/Avalonia.Controls.Maui/Extensions/MauiAppBuilderExtensions.cs
@@ -1,14 +1,18 @@
 using System.Collections;
 using Avalonia.Controls.Maui.Animations;
+using Avalonia.Controls.Maui.LifecycleEvents;
 using Avalonia.Controls.Maui.Services;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Maui;
 using Microsoft.Maui.Animations;
 using Microsoft.Maui.Hosting;
+using Microsoft.Maui.LifecycleEvents;
 
 public static class MauiAppBuilderExtensions
 {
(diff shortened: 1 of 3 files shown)


src/Avalonia.Controls.Maui/Extensions/MauiAppBuilderExtensions.cs
```diff
@@ -154,9 +158,39 @@ public static MauiAppBuilder UseAvaloniaApp(this MauiAppBuilder builder, bool us
...
+            events.AddWindows(avalonia =>
+            {
+                avalonia.OnLaunching((application, _) => RegisterResources(application));
+            });
+        });
+
+        return builder;
+    }
+
+    private static void RegisterResources(Avalonia.Application application)
```
jsuarezruiz, line 182: Works with AOT/trimming?

drasticactions: Yeah, that's what my screenshot shows. It's the Release version of the ControlGallery, served with `dotnet serve`

drasticactions: image

And here it is running with macOS with NativeAOT.

drasticactions:
Talking about this in Slack, let's just fix this in #69

PR #75, opened 2025-12-16
Unify several XAML control themes

maxkatz6:
Before this PR:

1. Some controls were used Style + Selector and Template setter (old way)
2. Other controls used ResourceDictionary + ControlTheme (new way), but referenced from C# code instead of single XAML style entrypoint (ControlStyles)
3. MauiSearchBar was a self-styled templated control
4. Some controls build their own template from C# code (CollectionView)
5. Minor: XAML files were public by default

After this PR:

1. Almost all controls now use ResourceDictionary + ControlTheme. Except controls that were built from the C# code - it would be more involving change.
2. Minor: XAML files are internal by default (ControlStyles is still public).
3. Added IServiceProvider to the `ControlStyles` ctor. It's only necessary, if this class was used from XAML, but better to add now than lock for a bug in the future.


drasticactions (approved):
1. Tested all of the samples with NativeAOT/Release mode and it worked fine, no warnings or errors.
2. _Thank you_ for this. I have zero idea which concepts are "correct" for a given case, and I've mostly been going through our various codebases to see how things are done, so that ends up with mixing and matching stuff depending on where it was cribbed from. And LLMs don't help here either 😉.

This helps me a ton, I very much appreciate it.

Issue #14, opened 2025-11-20
Implementing Obsolete Controls


drasticactions:
Controls like Frame and ListView are obsolete in MAUI for the versions we target. We need to decide if we're going to implement them anyway, as to allow people to port existing code to run with the new platforms, or only implement supported controls and have people upgrade.

It would make _far_ more sense to not implement obsolete controls from a support perspective, but those other controls are still used so asking for upgrades is a lift.

PR #76, opened 2025-12-16
Refactor TableViewHandler, cell handlers, added samples and tests 

jsuarezruiz:
This PR contain changes to complete the TableView control, refactor handler to use extension methods and also complete all the cells:
- EntryCell
- ImageCell
- SwitchCell
- TextCell
- ViewCell

Added support to all the properties, custom events, lifecycle events and visual details.

Added a complete set of samples in the Gallery:
tableviewhandler
tableviewhandler-moresamples
image

And tests:
image


drasticactions (approved):
I added a NoWarn for the obsolete control (https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/76/commits/e0c9ac17220c18d21ec71d5c45a337365c258e50) reference to keep ControlGallery green without warnings.

This code is good to me. Relating to https://github.com/AvaloniaUI/Avalonia.Controls.Maui/issues/14, IMO I think TableView makes sense to keep as a control, even though it's obsolete in MAUI. This isn't too much code to maintain and it's useful. For larger scale controls, like ListView, we should all sync up (ourselves and MAUI) about it, since that probably isn't worth doing (and I also don't want to, lol)

PR #79, opened 2025-12-17
Implement PageHandler background properties

jsuarezruiz:
Implement PageHandler background properties:
- Background
- BackgroundImageSource

pagehandler-background

Also added tests:
image


src/Avalonia.Controls.Maui/Extensions/PageExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+                            }
+                            
+                            if (bitmap != null)
+                            {
+                                await Threading.Dispatcher.UIThread.InvokeAsync(() =>
+                                {
+                                    platformView.Background = new global::Avalonia.Media.ImageBrush
+                                    {
+                                        Source = bitmap,
+                                        Stretch = Media.Stretch.Fill
```
drasticactions, line 53: Is this what MAUI does by default for an image page background? Are there options for setting it from a source where you can control this?

if it's not implemented in the other MAUI platforms that way and it does this, then that's fine, I just wanna make sure we're consistent (or if they're not consistent between platforms, we should tell them)

jsuarezruiz: This property don’t allow to control layout/rendering (Aspect, tiling, alignment). In MAUI, Android and Windows stretches in a way that can distort aspect ratio, while iOS tile/repeat: https://github.com/dotnet/maui/issues/27438
So, copied the more repeated behavior in the best possible way.

drasticactions (approved):
LGTM

PR #80, opened 2025-12-17
Implemented Alerts, ActionSheets, Prompts

jsuarezruiz:
Implemented Alerts, ActionSheets, Prompts and Page IsBusy (also inside the AlertManager).

alerts-light
alerts-dark


drasticactions (commented):
I think we need a broader conversation around Alerts. Creating Windows will work for Desktop, but it won't work for Browser, so we would either need something specific for Browser, or something that will work for all of them.

I had written it like that originally to see if it would work at all, but I think for RTM we need a broader approach.

jsuarezruiz:
> I think we need a broader conversation around Alerts. Creating Windows will work for Desktop, but it won't work for Browser, so we would either need something specific for Browser, or something that will work for all of them.
> 
> I had written it like that originally to see if it would work at all, but I think for RTM we need a broader approach.

Mnnn, right. One option could be, make alerts render inside the visual tree (a modal overlay layer) so the same mechanism works on Desktop + Mobile + Browser. thoughts?

jsuarezruiz:
> > I think we need a broader conversation around Alerts. Creating Windows will work for Desktop, but it won't work for Browser, so we would either need something specific for Browser, or something that will work for all of them.
> > I had written it like that originally to see if it would work at all, but I think for RTM we need a broader approach.
> 
> Mnnn, right. One option could be, make alerts render inside the visual tree (a modal overlay layer) so the same mechanism works on Desktop + Mobile + Browser. thoughts?

Applied changes to do the suggestions. Use an overlay layer to draw alerts and now, works on Desktop and Browser:
alerts-web

drasticactions (approved):
LGTM 👍

PR #82, opened 2025-12-19
Changes in DatePickerHandler, completed implementations, added samples and tests

jsuarezruiz:
Applied changes to the DatePickerHandler to complete implementations like CharacterSpacing and others and moved the logic to extension methods.

Added samples into the Gallery:
datepickerhandler

And several tests:
image

Also, updated the Status document.


drasticactions (approved):
LGTM. I added a CultureInfo set for the WASM app so we wouldn't need to pull in more custom fonts for now (Since CJK systems will show blank characters for the date glyphs, https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/82/commits/34df228d38ea0902a17be090b5ffbcab1eea1fa5
. I rather not blow up the size of the ControlGallery app unless we need to, and I think for this it's an example of how it works. We can always change it later too if we do want to include a CJK font.

PR #84, opened 2025-12-19
Improve Gallery

jsuarezruiz:
Changes applied to the Gallery:
- Added a SearchBar to filter samples.
- Added a Welcome page loaded initially.

Screenshot 2025-12-19 at 12 04 21


drasticactions (approved):
I updated the Dark Mode color banner to match the light mode one. I did play with adding a border around the TableView to try and match the Search Box as well so it wouldn't be cut off, 

Image

Not sure if that's better or not.

Issue #67, opened 2025-12-10
Throw WarningsAsErrors for Sample/Spotlight projects, make sure everything is trimmable


drasticactions:
Basically to prevent https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/65 from happening again.

PR #89, opened 2025-12-22
Refactor TimePickerHandler, added samples and tests

jsuarezruiz:
Applied changes to the TimePickerHandler to support more properties and moving mapping logic to public extension methods.

Added samples in the Gallery:
timepickerhandler

Several tests:
image

And update the Status.


jsuarezruiz:
Rebased and fixed conflicts!

drasticactions (approved):
There was a XamlC warning for the lacking DataType, so I added it. I need to do https://github.com/AvaloniaUI/Avalonia.Controls.Maui/issues/67 when I get back, but for now I turned on TreatWarningsAsErrors for ControlGallery to make sure we don't introduce any warnings to the ControlGallery app.

PR #90, opened 2025-12-22
Refactor and complete IndicatorViewHandler, added samples and tests

jsuarezruiz:
Completed pending stuff like using ItemsSource or templates. Also moved logic to public extension methods.

Added samples:
indicatorviewhandler

Added several tests:
image

And updated the Status.


src/Avalonia.Controls.Maui/Extensions/IndicatorViewExtensions.cs
```diff
@@ -0,0 +1,223 @@
...
+            try
+            {
+                // Create MAUI view from template
+                var mauiView = concreteView.IndicatorTemplate.CreateContent() as Microsoft.Maui.Controls.View;
+                if (mauiView == null)
+                    return CreateDefaultIndicator(platformView, isSelected);
+
+                // Set the binding context with anonymous type
+                mauiView.BindingContext = new { Index = index, IsSelected = isSelected };
+
```
drasticactions: The IndicatorView with DataTemplate has trimming issues:

```
ControlGallery net10.0 6 件の警告付きで成功しました (2.7 秒) → /Users/drasticactions/Developer/Work/Avalonia.Controls.Maui-Holder/Avalonia.Controls.Maui/samples/ControlGallery/ControlGallery/bin/Release/net10.0/ControlGallery.dll
    /Users/drasticactions/Developer/Work/Avalonia.Controls.Maui-Holder/Avalonia.Controls.Maui/samples/ControlGallery/ControlGallery/Pages/IndicatorViewPage.xaml(310,58): XamlC warning XC0022: Binding could be compiled to improve runtime performance if x:DataType is specified. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
    /Users/drasticactions/Developer/Work/Avalonia.Controls.Maui-Holder/Avalonia.Controls.Maui/samples/ControlGallery/ControlGallery/Pages/IndicatorViewPage.xaml(315,58): XamlC warning XC0022: Binding could be compiled to improve runtime performance if x:DataType is specified. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
    /Users/drasticactions/Developer/Work/Avalonia.Controls.Maui-Holder/Avalonia.Controls.Maui/samples/ControlGallery/ControlGallery/Pages/IndicatorViewPage.xaml(326,62): XamlC warning XC0022: Binding could be compiled to improve runtime performance if x:DataType is specified. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
    /Users/drasticactions/Developer/Work/Avalonia.Controls.Maui-Holder/Avalonia.Controls.Maui/samples/ControlGallery/ControlGallery/Pages/IndicatorViewPage.xaml(328,77): XamlC warning XC0022: Binding could be compiled to improve runtime performance if x:DataType is specified. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
    /Users/drasticactions/Developer/Work/Avalonia.Controls.Maui-Holder/Avalonia.Controls.Maui/samples/ControlGallery/ControlGallery/Pages/IndicatorViewPage.xaml(332,62): XamlC warning XC0022: Binding could be compiled to improve runtime performance if x:DataType is specified. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
    /Users/drasticactions/Developer/Work/Avalonia.Controls.Maui-Holder/Avalonia.Controls.Maui/samples/ControlGallery/ControlGallery/Pages/IndicatorViewPage.xaml(334,77): XamlC warning XC0022: Binding could be compiled to improve runtime performance if x:DataType is specified. See https://learn.microsoft.com/dotnet/maui/fundamentals/data-binding/compiled-bindings for more information.
```

There isn't a way to set a DataType for this as it's internal to this, so it won't work correctly as far as I can tell.

drasticactions (changes requested):

src/Avalonia.Controls.Maui/Extensions/IndicatorViewExtensions.cs
jsuarezruiz: Updated the sample to use: `<DataTemplate x:DataType="x:Null">`
I think with this, now suppressed the compiled binding warnings.

drasticactions: I think that masks the issue, it doesn't fix it. The object needs to be exposed so it can properly be bound to. The DataType is for both Compiled Bindings and to allow for type ahead to work, so having it as null, you would not know what the types were in the template, and it's not obvious to a user you would need to set it to null for it to work. I think this also implies that this is the _only_ datatype supported for this template. If you tried to bind another object type to this view, it would fail since it's overridden by this anonymous type.

This looks like a bug to me.

How does MAUI normally handle this for this view? If it does the same anonymous type stuff then I think we should file a bug in MAUI to have a public type that can be bound to, since it would have the same issue here.

jsuarezruiz: Could create a class like:

```
namespace Avalonia.Controls.Maui.Controls;

public class IndicatorItem
{
    public int Index { get; }
    
    public bool IsSelected { get; }
    
    public IndicatorItem(int index, bool isSelected)
    {
        Index = index;
        IsSelected = isSelected;
    }
}
```

Then update our implementation:

`mauiView.BindingContext = new IndicatorItem(index, isSelected);`

And, allow to do:

`<DataTemplate x:DataType="controls:IndicatorItem">`

But, I think this should be fixed at MAUI level. What do you think if we:
- Create a request in the MAUI repo and an issue to remember this here
- Merge this PR to get everything except this for now, and return to it once the MAUI Team fix it
?

drasticactions: Having the hard class type to bind to and creating a MAUI issue to fix it there for .NET 11 would be best I think. If someone did bind to the controls:IndicatorItem type and we did change it inside the preview window, that IMO would be fine as a change, it's a minor break and not hard to fix.

jsuarezruiz:
Need to be redone with https://github.com/AvaloniaUI/Avalonia/pull/20660, close for now and will reopen.

PR #91, opened 2025-12-23
Refactor TabbedPageHandler, completed functionality, added samples and tests

jsuarezruiz:
Applied changes to the TabbedPageHandler completing functionality like ItemsSource, Icon, colors and others. All the logic moved to public extension methods.

Added samples to the Gallery covering all the possibilities:
tabbedpagehandler

Several tests:
image

And updated the Status.


drasticactions (approved):
Playing around with the Fluent Theme and this, I think we may need to adjust the Gallery defaults whenever we switch it,

Image
Image

But the defaults work fine, so I think this should be good for now.

PR #94, opened 2025-12-29
Add FrameHandler, samples and tests

jsuarezruiz:
Completed FrameHandler implementing all the mapping logic in extension methods.

_NOTE: Frame is deprecated in .NET MAUI 9+. But is a case similar to TableView for example but, Frame extensibility used in Xamarin.Forms. Thinking in future pending ports from Xamarin, have sense to have it._

Added samples:
framehandler

Several tests:
image

Updated also Status.


drasticactions (approved):
Ran of WASM, worked fine. No Trimming errors. LGTM!

PR #95, opened 2025-12-29
Created Maps project, handler, added samples and tests

jsuarezruiz:
Created the Avalonia.Controls.Maui.Map project where the MapHandler is implemented.

Added samples:
maphandler02

And several tests:
image

Also, updated the Status file. With the changes we have almost everything implemented (except the `IsTrafficEnabled` property that would require an API Key to support it).


src/Avalonia.Controls.Maui.Maps/Avalonia.Controls.Maui.Maps.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <AssemblyOriginatorKeyFile>$(BuildDirectory)/avalonia.snk</AssemblyOriginatorKeyFile>
+    <PackageDescription>Avalonia.Controls.Maui.Maps provides map support for .NET MAUI apps using Mapsui.Avalonia.</PackageDescription>
+    <SignAssembly>true</SignAssembly>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+    
+  <ItemGroup>
+    <PackageReference Include="Avalonia" Version="$(AvaloniaVersion)" />
+    <PackageReference Include="Mapsui.Avalonia" Version="5.0.0" />
+    <PackageReference Include="Mapsui.Nts" Version="5.0.0" />
```
drasticactions: This would make us dependent on a third-party library for the Avalonia implementation. I'm not wholesale against that, but I feel for us to implement a maps control "in the box", we would need to have an implementation we directly support and maintain, which would be a huge cost sink.

It could be better to have this library be an "interface" library without an implementation, so any of the existing Avalonia mapping libraries could support it. Then, for ControlGallery, it could have the Mapsui.Avalonia implementation there to show as an example of how to do one, or we could ship that implementation as a "Community" library, bound to this reference interface.

jsuarezruiz: I completely understand the concern about depending on a third-party library for Avalonia but maintaining a full maps control ourselves would indeed be a significant effort.

I like the idea of having the library act as a “reference interface” without an implementation. That would keep it flexible and let others plug in their mapping library. 

Will apply changes.

jsuarezruiz: Separated the Maps library into an interface-only library and a Mapsui implementation library.

The idea is that `IMapHandler` enables third-party map providers to implement their own backends.

src/Avalonia.Controls.Maui.Maps.Mapsui/Handlers/MapsuiMapHandler.cs
```diff
@@ -0,0 +1,426 @@
...
+                else
+                {
+                    var request = new GeolocationRequest(GeolocationAccuracy.Medium, TimeSpan.FromSeconds(10));
+                    location = await Geolocation.Default.GetLocationAsync(request);
+                }
+                
+                if (location != null)
+                {
+                     UpdateUserLocation(handler, mapControl, locationLayer, location);
+                }
```
drasticactions: Running the sample on Coffee shop Wifi, when I clicked on the map control I got a UI hang and

`IP Geolocation failed: The request was canceled due to the configured HttpClient.Timeout of 5 seconds elapsing.`

in the output. It eventually dropped trying and let me though. Running it a second time, it worked. This is happening on the UI Thread, does it have to?

We should verify that the methods called here only touch the UI Thread if/when they need to.

drasticactions (changes requested):
I'm cool with the API. I think we need a wider Avalonia-level talk about if we should have a Mapsui binding as well but apart from the UI Thread issues I think I hit below, the implementation should be fine.

src/Avalonia.Controls.Maui.Maps.Mapsui/Handlers/MapsuiMapHandler.cs
jsuarezruiz: Good catch. This is related with the IP Geolocation logic. I have implemented some changes:
Key improvements:
- Background Offloading: Geolocation now runs on a background thread.
- Cancellation support: Added cancellation handling when toggling "Show User Location" or disconnecting the map.
- Optimized Async: Used ConfigureAwait(false) for internal tasks and properly handled timeouts.

Tried with WiFi and mobile data, and I think should fix it.

drasticactions:
You can't use gestures within the maps if they are within a scroll view. If you do the page will continue to scroll. The Collection/List views which have nested scroll views don't have that issue (It does scroll the page if you hit the bottom of a list, which is a little weird, but I need to check with Avalonia itself for what happens there)


https://github.com/user-attachments/assets/25e969aa-a110-4caf-80aa-404319f8a2a0

drasticactions (approved):

PR #100, opened 2026-01-08
Refactor EntryHandler to complete it, adding samples and tests

jsuarezruiz:
Applied changes to the **Entry** handler to complete the functionality and moving mapping logic to extension methods.

Added samples in the Gallery:
entryhandler1
entryhandler2

Added several tests:
image

And updated the Status.


drasticactions:
~~@jsuarezruiz I rebased this on main, please validate it so it works as you expect.~~

I lied, you know more of what's needed here. Please rebase this and I'll merge it. Sorry!

src/Avalonia.Controls.Maui/Extensions/EntryExtensions.cs
```diff
@@ -18,6 +19,7 @@ public static class EntryExtensions
     /// </summary>
     /// <param name="textBox">The Avalonia TextBox.</param>
     /// <param name="entry">The .NET MAUI Entry.</param>
```
drasticactions, line 21: We can't suppress this. I'm unsure which elements would require needing this reflection to exist where we get need to get it this way.

https://github.com/dotnet/maui/blob/9c0887b8c343a1a98762acfccb64190355e175ee/src/Controls/src/Core/InputView/InputView.cs#L178-L182

https://github.com/dotnet/maui/blob/9c0887b8c343a1a98762acfccb64190355e175ee/src/Controls/src/Core/ITextElement.cs#L48

Element uses InputView, which implements ITextElement, that contains TextTransform. We don't need reflection for this.

samples/ControlGallery/ControlGallery/Pages/EntryPage.xaml.cs
```diff
@@ -1,5 +1,9 @@
+using System.Diagnostics.CodeAnalysis;
+
 namespace ControlGallery.Pages
 {
+    [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "MAUI XAML loader usage is unavoidable here.")]
```
drasticactions, line 5: I removed both of these lines and built with release mode, and it worked fine. I'm not sure if there was something else on this page previously that required this. If there was, it's gone now AFAIK, so this can be removed. But in general, we should never have suppressed entries for AOT and trimming, either in ControlGallery or the library itself. ControlGallery is built with NativeAOT, so it would fail when we publish the Linux builds if it did break on trimming, same for the library.

drasticactions (changes requested):
These suppress the AOT issues, it doesn't fix them. We need to make sure trimming is done for both ControlGallery and the library itself. If anything did require us to need reflection in MAUI, then we need to tell them ASAP because that's a problem.

drasticactions:
I believe EntryStub also has to implement ITextElement, https://github.com/AvaloniaUI/Avalonia.Controls.Maui/blob/entryhandler/tests/Avalonia.Controls.Maui.Tests/Stubs/EntryStub.cs

https://github.com/dotnet/maui/blob/9c0887b8c343a1a98762acfccb64190355e175ee/src/Controls/src/Core/Entry/Entry.cs#L12

Maui Element is an InputView, and the InputView implements ITextElement. The stub implements IEntry.

jsuarezruiz:
> I believe EntryStub also has to implement ITextElement, https://github.com/AvaloniaUI/Avalonia.Controls.Maui/blob/entryhandler/tests/Avalonia.Controls.Maui.Tests/Stubs/EntryStub.cs
> 
> https://github.com/dotnet/maui/blob/9c0887b8c343a1a98762acfccb64190355e175ee/src/Controls/src/Core/Entry/Entry.cs#L12
> 
> Maui Element is an InputView, and the InputView implements ITextElement. The stub implements IEntry.

Right, already implemented and fixed the tests.

drasticactions (approved):
LGTM, thanks! 👍

PR #104, opened 2026-01-14
Throw WarningsAsErrors for all the things, build the sample apps

### LINKED ISSUE #67: Throw WarningsAsErrors for Sample/Spotlight projects, make sure everything is trimmable

drasticactions:
Fixes #67 

Screenshot From 2026-01-14 17-46-29

This PR adds `TreatWarningsAsErrors` for all projects except for the Sandbox apps, and builds and bundles the sample and showcase projects as artifacts, with Linux x64 AppImages for the AOT versions. This is to ensure that will can always build with NativeAOT and so we can test the builds after the fact.

The biggest AOT change was for [MauiTableView](https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/104/changes#diff-8954f9ba5207d3258141b284bb4414049e4e7ceeccb71f9d404171d6016c45f9). The rest of it should be mostly Null warnings and some light trimming stuff that had to be adjusted

### DIFF
--- a/.github/workflows/build-desktop-samples.yml
+++ b/.github/workflows/build-desktop-samples.yml
@@ -9,30 +9,8 @@ on:
 
 jobs:
   build:
-    strategy:
-      fail-fast: false
-      matrix:
-        app:
-          - name: ControlGallery.Desktop
-            project: samples/ControlGallery/ControlGallery.Desktop/ControlGallery.Desktop.csproj
-          - name: MauiPlanets.Desktop
-            project: showcase/MauiPlanets/MauiPlanets.Desktop/MauiPlanets.Desktop.csproj
-          - name: AlohaKit.Gallery.Desktop
-            project: showcase/AlohaKit.Gallery/AlohaKit.Gallery.Desktop/AlohaKit.Gallery.Desktop.csproj
-          - name: 2048Game.Desktop
-            project: showcase/2048Game/2048Game.Desktop/2048Game.Desktop.csproj
-        build-type:
-          - name: Release
-            publish-aot: false
-            rid: ''
-            os: ubuntu-latest
-          - name: Release-AOT
-            publish-aot: true
-            rid: linux-x64
-            os: ubuntu-latest
-
-    runs-on: ${{ matrix.build-type.os }}
-    name: ${{ matrix.app.name }} (${{ matrix.build-type.name }})
+    runs-on: ubuntu-latest
+    name: Build All Desktop Samples
 
     steps:
       - name: Checkout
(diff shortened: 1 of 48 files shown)


.github/workflows/build-desktop-samples.yml
```diff
@@ -0,0 +1,150 @@
...
+          dotnet publish ${{ matrix.app.project }} \
+            -c Release \
+            -r ${{ matrix.build-type.rid }} \
+            -o ./artifacts/${{ matrix.app.name }}-${{ matrix.build-type.name }} \
+            /p:PublishAOT=true
+
+      - name: Create AppImage (AOT)
+        if: matrix.build-type.publish-aot
+        run: |
+          # Download appimagetool
```
maxkatz6, line 62: I would suggest Parcel here...but Parcel only provides DEB/RPM

.github/workflows/build-desktop-samples.yml
```diff
@@ -0,0 +1,150 @@
...
+          # Make the AppImage executable
+          chmod +x "./artifacts/$APP_NAME-x86_64.AppImage"
+
+      - name: Upload AppImage
+        if: matrix.build-type.publish-aot
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.app.name }}-AppImage
+          path: ./artifacts/${{ matrix.app.name }}-x86_64.AppImage
+          retention-days: 14
```
maxkatz6, line 122: With private repositories, I would keep retention lower. We essentially pay for every megabyte uploaded at a time.

.github/workflows/build-desktop-samples.yml
drasticactions: Maybe gives an idea for the future ;)

Yeah, I figured this would be easier for now for people to just run out of the box, although I guess we could also just have the exe too.

.github/workflows/build-desktop-samples.yml
drasticactions: Yeah, we don't need three versions for every build: Changed with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/104/commits/22cecdd560edca5fdf290b6c28913cb238d075a9

e. This also will be public at some point, so for the rate I think this should be okay, but I'm also fine with just doing the build and not publishing for everything, we just need to make sure it doesn't break.

jsuarezruiz (approved):

PR #106, opened 2026-01-19
Add existing MAUI Native TFMs to Avalonia.Controls.Maui

drasticactions:
This PR adds the existing MAUI TFMs (iOS, maccatalyst, android, windows) to the Avalonia.Controls.Maui nuget and implements the needed interfaces to let it build. This will let us enable Avalonia.Controls.Maui on existing MAUI UI platforms beyond Desktop and WASM. 

In MAUI Native, AlertManager and GestureManager call on native code that will call our code to fail. For iOS and Catalyst, as long as you don't invoke it, it "works" but on Android it will fail out of the box on GestureManager. This ["magic build"](https://github.com/drasticactions/maui/compare/dotnet%3Amaui%3Amain...magic-build) turns it off to let it past it.


.github/workflows/main.yml
```diff
@@ -46,10 +46,14 @@ jobs:
   build:
     needs: check-changes
     if: needs.check-changes.outputs.test-changes == 'true' || needs.check-changes.outputs.maui-changes == 'true' || needs.check-changes.outputs.build-changes == 'true'
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
```
jsuarezruiz, line 49: The change from `ubuntu-latest` to `windows-latest` is to support all TFMs, right?
Have any impact in costs? Could maintain Linux CI capability by conditionally excluding iOS/MacCatalyst TFMs, but this is simpler.

samples/ControlGallery/ControlGallery.AvaloniaMaui/ControlGallery.AvaloniaMaui.csproj
```diff
@@ -33,25 +33,19 @@
 
 		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">15.0</SupportedOSPlatformVersion>
 		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'">15.0</SupportedOSPlatformVersion>
-		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">21.0</SupportedOSPlatformVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">23.0</SupportedOSPlatformVersion>
```
jsuarezruiz: The sample project uses API 23 while Avalonia.Controls.Maui.csproj uses API 21. Seems MAUI minimum supported API Level is 24: https://learn.microsoft.com/en-us/dotnet/maui/supported-platforms?view=net-maui-10.0

tests/Avalonia.Controls.Maui.Tests/Avalonia.Controls.Maui.Tests.csproj
```diff
@@ -7,19 +7,23 @@
...
+    <!-- Disable xUnit v3 analyzer rule that requires TestContext.Current.CancellationToken -->
+    <NoWarn>$(NoWarn);xUnit1051</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="coverlet.collector" Version="6.0.4" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageReference Include="NSubstitute" Version="5.3.0" />
-    <PackageReference Include="xunit" Version="2.9.3" />
+    <PackageReference Include="xunit.v3" Version="3.2.1" />
```
jsuarezruiz, line 18: No warnings, changes with the upgrade, right?

.github/workflows/main.yml
drasticactions: As far as I'm aware you can't build the iOS/Mac assemblies on Linux at all, because the workloads don't exist. Likewise with the Windows ones as far as I can tell (even with Windows Targeting enabled, it also failed when building it for the nuget).

In terms of cost, it's only the build that needs to change, not the tests nor the samples, so AFAIK the cost shouldn't be that much higher. If it was on macOS then it would be bad.

tests/Avalonia.Controls.Maui.Tests/Avalonia.Controls.Maui.Tests.csproj
drasticactions: There was none before your render tests PR, now it doesn't work at all, lol. So I need to fix it.

drasticactions: Fixed with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/106/commits/989e995cacda653daf6bf428eed9a23206ab55c2

samples/ControlGallery/ControlGallery.AvaloniaMaui/ControlGallery.AvaloniaMaui.csproj
drasticactions: I synced the versions with https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/106/commits/8efc308ab5b710141ff6d667ff9b7dec210c93c0. We'll need to verify what Avalonia itself targets for v12, since it will be whatever version ends up highest.

jsuarezruiz (approved):

PR #107, opened 2026-01-20
Added Render Tests

jsuarezruiz:
Added a Render tests project. 

The tests use MAUI code and a compare images method to validate the rendering. Example:

```
[AvaloniaFact]
public async Task Render_Button_Default()
{
    var button = new Microsoft.Maui.Controls.Button
    {
        Text = "Click Me",
        BackgroundColor = Color.FromRgb(0, 120, 215),
        TextColor = Color.FromRgb(255, 255, 255),
        WidthRequest = 120,
        HeightRequest = 40
    };

    await RenderToFile(button);
    CompareImages();
}
```

The reference snapshots are available in the Assets folder and generated the first time that pass the test.

image

To avoid visual regressions, is a nice addition to the unit tests project.


src/Avalonia.Controls.Maui/Fonts/FontManager.cs
```diff
@@ -30,7 +30,7 @@ public FontManager(IFontRegistrar fontRegistrar, IServiceProvider? serviceProvid
     /// Gets the default font family for the application.
     /// </summary>
     public global::Avalonia.Media.FontFamily DefaultFontFamily =>
-        _defaultFontFamily ??= global::Avalonia.Media.FontFamily.Default;
+        _defaultFontFamily ??= new global::Avalonia.Media.FontFamily("Inter");
```
drasticactions: "Inter" is included in the default Avalonia template app, but AFAIK it does not mean that you can assume it is included with an application in general.

@Gillibald What do you think about this?

Gillibald: It should be using FontFamily.Default. Inter is not generally available as a system font. The user can always adjust the default via FontManagerOptions if required.

drasticactions: @jsuarezruiz Please revert this change. If the render tests need a default font, I think OpenSans should be okay. The MAUI template include it, and we need the render tests to be consistent, I don't think that change needs to be in the root code.

Gillibald: Render tests should always use some embedded font. Either construct the FontFamily explicitly: https://github.com/AvaloniaUI/Avalonia/blob/master/tests/Avalonia.Skia.UnitTests/Media/FontManagerTests.cs#L60

Or change the system font collection: https://github.com/AvaloniaUI/Avalonia/blob/master/tests/Avalonia.Skia.UnitTests/Media/FontManagerTests.cs#L275

FontManagerOptions can be applied like so:
https://github.com/AvaloniaUI/Avalonia/blob/master/tests/Avalonia.Skia.UnitTests/Media/FontManagerTests.cs#L375

The last test also shows how to define some mapping

drasticactions:
Is it possible to upload the generated render images as part of the test results? That way we can compare what they are with the originals, since some of them based on the failures seem very close within tolerance, and we can see what they are when run on CI compared to our local machines.

drasticactions:
For the DatePicker, it changes depending on the users Locale,

DatePickerRenderTests_Render_DatePicker diff

so it will fail for me because my computer is in ja-JP. The tests should probably be run with either an en-US local or a neutral culture.

jsuarezruiz:
> Is it possible to upload the generated render images as part of the test results? That way we can compare what they are with the originals, since some of them based on the failures seem very close within tolerance, and we can see what they are when run on CI compared to our local machines.

Yes, added a new step to upload images:
image

In addition to upload the created images, create a .diff one comparing between reference and result:
image

- Green bar = Expected
- Blue bar = Actual
- Red bar = Diff
    
Also, draw in red the differences in the diff.

Having small issues with fonts!

tests/Avalonia.Controls.Maui.RenderTests/App.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Avalonia.Controls.Maui.RenderTests;
+
+public class App : Application
+{
+    public override void Initialize()
+    {
+        RequestedThemeVariant = ThemeVariant.Light;
+        Styles.Add(new FluentTheme());
+        Styles.Add(new Avalonia.Themes.Simple.SimpleTheme());
```
drasticactions: It should be either Fluent or Simple. Fluent is the one used by default in the Avalonia template so that should be the one to test with.

jsuarezruiz:
> For the DatePicker, it changes depending on the users Locale,
> 
> DatePickerRenderTests_Render_DatePicker diff
> so it will fail for me because my computer is in ja-JP. The tests should probably be run with either an en-US local or a neutral culture.

Good catch. Added CultureInfo.InvariantCulture in EnsureHandlerCreated(). Now tests will use invariant culture regardless of local machine settings (ja-JP, en-US, etc.).

tests/Avalonia.Controls.Maui.RenderTests/App.cs
drasticactions: It should be either Fluent or Simple. Fluent is the one used by default in the Avalonia template so that should be the one to test with.

drasticactions (changes requested):
Once this is fixed, I think this PR is good to go.

tests/Avalonia.Controls.Maui.RenderTests/App.cs
jsuarezruiz: Fixed!

drasticactions (approved):

PR #108, opened 2026-01-21
Add Compatibility project

### LINKED ISSUE #103: Compatibility library
We currently have implementations for several MAUI controls that are obsolete at the platform level (e.g. `ListView`, `Frame`, `TableView`). While these controls are still widely used, especially by developers porting apps from Xamarin.Forms, their inclusion in the core libraries creates a long-term maintenance burden and raises questions about alignment with future .NET releases (notably .NET 11).

There is also a risk that improved implementations of obsolete controls (e.g. `ListView`) could divert users away from modern, recommended controls such as `CollectionView`, which would be counterproductive.

At the same time, removing these controls outright would make porting legacy apps harder and increase friction for existing users.

## Proposal
Create a **separate Compatibility library/package** that houses handlers and views for obsolete or legacy controls, decoupling them from the main supported surface area while still offering a migration path for users who need them.

This library would:
- Contain implementations for controls such as:
  - `ListView`
  - `Frame`
  - `TableView`
  - (and any other obsolete/legacy controls as identified)
- Be clearly documented as **compatibility / legacy support**
- Allow us to:
  - Keep the core libraries focused on modern, supported controls
  - Reduce long-term maintenance pressure
  - Align better with future .NET versions (e.g. .NET 11)
  - Transition ownership to a community-maintained package in the future (e.g. MAUI Community Toolkit)

## Tasks / Acceptance Criteria
- [ ] Create a new Compatibility library/package for obsolete controls  
- [ ] Move existing implementations of `ListView`, `Frame`, `TableView`, etc. into this library  
- [ ] Update internal apps (e.g. ControlGallery) to reference the Compatibility library instead of core  
- [ ] Ensure clear documentation and messaging that these controls are legacy/compatibility-only  

## Notes
This approach allows us to support current users and porting scenarios without committing to indefinite maintenance of obsolete APIs in the core stack. It also gives the community an opportunity to extend or maintain these controls independently if there is sufficient demand.

jsuarezruiz:
Added Compatibility project containing deprecated control handlers: Frame, ListView and TableView.

To use this handlers require to add a reference to the new **Avalonia.Controls.Maui.Compatibility** library and use the **UseAvaloniaCompatibility** extension method.

Fixes https://github.com/AvaloniaUI/Avalonia.Controls.Maui/issues/103


drasticactions (approved):
LGTM 👍

PR #110, opened 2026-01-30
Refactor and complete Shell handlers, added samples and tests

jsuarezruiz:
Refactor and complete Shell handlers: Shell, ShellContent, ShellItem, SearchHandler etc. Continue the previous works fixing some parts (parenting, propagating some changes etc) and implement pending parts: SearchHandler, navigation animations, colors etc.

Added two samples in the Gallery. The first one is Xaminals from https://github.com/dotnet/maui-samples/tree/main/10.0/Fundamentals/Shell/Xaminals, the official Shell sample used in the docs.
shell-xaminals

And created a more complete and complex gallery to validate all the core scenarios around Flyout, navigation, Tabs etc.
shell-playground-1
shell-playground-2
shell-playground-3
shell-playground-4
shell-playground-5
shell-playground-6
shell-playground-7
shell-playground-8

Added several tests for each case.

And updated Status.


samples/ControlGallery/ControlGallery/Pages/Shell/Playground/ShellSearchPage.xaml.cs
```diff
@@ -0,0 +1,235 @@
...
+        private void OnClearSearch(object sender, EventArgs e)
+        {
+            var handler = Shell.GetSearchHandler(this);
+            if (handler != null)
+            {
+                handler.Query = string.Empty;
+            }
+        }
+
+        [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Sample code with known properties")]
```
drasticactions: Is this true? If you ran this with trimming, would this code still run?

src/Avalonia.Controls.Maui/Controls/Shell/ShellSearchControl.cs
```diff
@@ -0,0 +1,754 @@
...
+
+            var itemsSource = _mauiSearchHandler.ItemsSource;
+            _resultsList.ItemsSource = itemsSource;
+            
+            UpdateShowsResults();
+        }
+
+        [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "DisplayMemberName requires reflection.")]
+        [UnconditionalSuppressMessage("Trimming", "IL2075", Justification = "DisplayMemberName requires reflection.")]
+        [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "DisplayMemberName requires reflection.")]
```
drasticactions: Suppressed Error

drasticactions (changes requested):
There are UnconditionalSuppressMessage messages hiding trimming and AOT errors. The Search page needs to be verified in NativeAOT mode, the ShellSearchControl ones need to be fixed; they can't be suppressed. We either need to find a publicly available value to bind to or disable this code and ask MAUI to open it.

drasticactions:
@jsuarezruiz Looking at this closer

https://github.com/dotnet/maui/blob/c1e543e1f116911ffb934e60c79bb501ea1814e3/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.targets#L300-L301

https://github.com/dotnet/maui/blob/c1e543e1f116911ffb934e60c79bb501ea1814e3/src/Controls/src/Core/TrimmerConstants.cs#L14

https://github.com/dotnet/maui/blob/c1e543e1f116911ffb934e60c79bb501ea1814e3/src/Controls/src/Core/Shell/SearchHandler.cs#L518-L524

DisplayMemberName is obsolete and not trimmable. ItemTemplate is preferred so IMO this should be the only use case we support. Otherwise we would need the suppressions and ways to split up the code to allow non-trimmable stuff and I rather not do that at all costs if we can.

jsuarezruiz:
> DisplayMemberName

Applied changes:
  - Dropped using System.Diagnostics.CodeAnalysis
  - Removed the 3 `[UnconditionalSuppressMessage]` attributes from` UpdateItemTemplate()`
  - Removed the entire `GetPropertyValue()` reflection helper
  - Simplified `UpdateItemTemplate()`: when no ItemTemplate is set, the fallback now just calls `item?.ToString()`, no reflection, fully trimmable/AOT-safe

drasticactions (approved):
LGTM 👍

PR #111, opened 2026-01-30
Improve Gallery: Selected sample indicator, etc

jsuarezruiz:
Changes:
- Selection Highlighting: Replaced `TextCell` with `ViewCell` in the main menu to provide visual feedback for the currently selected sample.
- Clean up Polish: Removed "Control" word from samples titles because in MAUI the terminology use Views, not Controls.

selection-light
selection-dark


drasticactions (approved):
Not entirely sure rewording most of the comments to be slightly different is _really_ needed but I can live, lol.

This is fine, but it may be better once https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/110 lands for us to try and rewrite this to be Shell instead, instead of using an obsolete control. That would probably encourage more people to use it, which we shouldn't be trying to do anyway.

PR #117, opened 2026-02-12
.NET 11 bump

drasticactions:
This PR bumps up the runtime to .NET 11. We have decided that our current course of action will be to:

- Continue going ahead and targeting .NET 11 as the main platform for release
- Leave the .NET 10 version flags in CLI use if you want to build against them locally for testing

I also added some basic README docs and updated it so the build version baseline should be based on the MAUI version. I need to think about it more, but I think the right thing to do is stick with the MAUI Version we target as the one we release with, since if we use the Avalonia version (or do some other versioning) it will be very hard to know what MAUI version that the package targets. That MAUI target matters more, so that should be the versioning we align to.


jsuarezruiz:
Do we want it in a net11 branch maybe? Build is passing!

drasticactions:
> Do we want it in a net11 branch maybe? Build is passing!

I need to sleep on it but I think the right thing to do is for main to be net11

- The first release is targeting MAUI net11.0 branched code, any change we need in MAUI needs to be in their net11 release branch. So we should keep that as the target.
- That does mean needing to install .NET 11 previews/workloads/etc which is painful...
- But this would be representative of what we are going to ship, and while their main is still AFAIK net10 based, I don't want to rely on code always landing there and then getting pushed to the net11 branch, especially when it does flip.
- Also, we are not going to release any net10 builds, it will only be net11 to target those MAUI builds.

At least for me running it locally, I haven't had issues running the net11 SDK P1 and its workloads in this codebase, and it's been working fine for net10 and below too. The other Avalonia projects include global.json (this probably should too) so it will use the older SDKs for those. Could probably also set up a local `.dotnet` version for this too so it doesn't need to be installed globally. I'll see if that'll work.

PR #120, opened 2026-02-14
Use Fluent Theme for samples/showcase apps

drasticactions:
Cleans up the samples and showcase apps to use the Fluent Theme by default, rather than simple. Leaving the tests to simple since at least right now it shouldn't make a difference.

Also cleaned up the default button font for ControlGallery so it uses Open Sans by default.


jsuarezruiz (commented):
Are there any noticeable visual change in some case?

drasticactions:
https://github.com/user-attachments/assets/a096da88-50ac-4b89-9b2f-2e508f51ab35

For the most part, since the controls are handled by whatever the MAUI style is, they won't change. The biggest ones are ones that map directly to Avalonia and use its themes, so like tab bars and stuff. Some of them looked jank with the other theme too, so I think it's more we need to think about how these controls interact against the MAUI themes, since the colors and stuff might not map one-to-one with what a user would see in the Native apps.

PR #121, opened 2026-02-18
Bump to Avalonia 12.0 preview 1

drasticactions:
Bump to Avalonia 12.0.0-preview1, and the net10.0 MAUI nuget to 10.0.40

### DIFF
--- a/src/Avalonia.Controls.Maui/Controls/MauiComboBox/MauiComboBox.axaml
+++ b/src/Avalonia.Controls.Maui/Controls/MauiComboBox/MauiComboBox.axaml
@@ -169,7 +169,7 @@
                        Foreground="{TemplateBinding Foreground}"
                        Background="Transparent"
                        Text="{TemplateBinding Text, Mode=TwoWay}"
-                       Watermark="{TemplateBinding PlaceholderText}"
+                       PlaceholderText="{TemplateBinding PlaceholderText}"
                        BorderThickness="0"
                        IsVisible="{TemplateBinding IsEditable}">
                 <TextBox.Resources>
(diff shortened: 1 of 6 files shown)


src/Avalonia.Controls.Maui/Controls/MauiComboBox/MauiComboBox.axaml
```diff
@@ -169,7 +169,7 @@
                        Foreground="{TemplateBinding Foreground}"
                        Background="Transparent"
                        Text="{TemplateBinding Text, Mode=TwoWay}"
-                       Watermark="{TemplateBinding PlaceholderText}"
+                       PlaceholderText="{TemplateBinding PlaceholderText}"
```
drasticactions, line 172: Watermark is obsolete, so changed to PlaceholderText

src/Avalonia.Controls.Maui/Controls/MauiEntry/MauiEntry.cs
```diff
@@ -24,21 +24,6 @@ public ClearButtonVisibility ClearButtonVisibility
         set => SetValue(ClearButtonVisibilityProperty, value);
     }
 
-    /// <summary>
-    /// Defines the <see cref="PlaceholderForeground"/> property.
```
drasticactions, line 28: Placeholder is now in TextBox, so this is removed.

src/Avalonia.Controls.Maui/Controls/MauiComboBox/MauiComboBox.axaml
jsuarezruiz: Nice, this is something that will happen with more properties even entire controls like `NavigationPage` but are good news. More possibilities in the core, less specific code to maintain here.

jsuarezruiz (approved):

PR #126, opened 2026-02-20
Map Background Color for BoxView

### LINKED ISSUE #123: BoxView.Background (And probably other mappers) need to be mapped.
BoxView.Background isn't mapped, so colors won't appear.

drasticactions:
Fixes #123 

This updates `ViewExtensions` to handle `Border` controls to update the background, and updates the `BoxView` handler to not clear the background value if `Background` is already set.

The way it _should_ work is:

- Color takes precedence, if you set that it will override everything.
- Background is next
- BackgroundColor is last.
 
スクリーンショット 2026-02-20 14 44 18


jsuarezruiz (approved):

jsuarezruiz (changes requested):
`Null Color Clears Background` test is failing.

drasticactions:
@jsuarezruiz Turns out that `BoxView.Background` is never null. I needed to adjust the logic a bit to make it work so that it checks if the the BoxView background exists and is empty, _then_ it's safe to clear the border background. So that should fix Color/BackgroundColor/Background

PR #132, opened 2026-02-24
Add Aloha AI (and fix a bunch of issues along the way)

drasticactions:


drasticactions:
The Sqlite IL issues are fine, it's in the library and not being called by the app. The Homepage one is legit, but didn't happen on Mac, so I need to fix that. Once that is done, it should be fine.

PR #134, opened 2026-02-25
Add the MAUI Repo Sample App (With fixes and tests to support it)

drasticactions:
スクリーンショット 2026-02-25 18 11 29


drasticactions:
For now, this won't be in the CI for builds because of various trimming stuff that needs to be dealt with. Since it was never intended for AOT (being it's the MAUI test bed sample) that's fine. I'll update it later. 

The fixes and new tests are what really matter, and those worked.

PR #144, opened 2026-03-02
Disable Maps temporary

jsuarezruiz:
Until Mapsui  https://github.com/Mapsui/Mapsui updates their Avalonia references, just avoid to launch the Maps sample from the Gallery.

The crash comes from Mapsui.Avalonia 5.1.0-beta.2 still targeting Avalonia 11, but the project is already on 12.0.999-alpha. When the Maps assembly is loaded, Mapsui's MapControl tries to resolve Avalonia 11 APIs that no longer exist.


drasticactions (approved):
Yeah, the code is fine (as in, the Maps code in this repo is fine). https://github.com/Mapsui/Mapsui/issues/3299 needs to be resolved.

PR #147, opened 2026-03-03
Added Config and Setup documentation   

jsuarezruiz:
Changes:
  - Adds `docs/config-and-setup.md` covering `MauiAppBuilder` extension methods (`UseAvaloniaApp`, `UseAvaloniaGraphics`, `UseAvaloniaCompatibility`, `UseAvaloniaEssentials`), font registration, image and raw asset handling, and custom handler registration.

@drasticactions Is the correct place where you plan to put the documentation?

### DIFF
--- a/docs/config-and-setup.md
+++ b/docs/config-and-setup.md
@@ -1,14 +1,34 @@
 # Configuration and setup
 
-`Avalonia.Controls.Maui` replaces the .NET MAUI native rendering pipeline with Avalonia. Before your app can use Avalonia-drawn controls, fonts, or assets, you need to register the appropriate services in `MauiProgram.cs`. This is done by chaining a set of extension methods onto the `MauiAppBuilder` object in the `CreateMauiApp` method.
+`Avalonia.Controls.Maui` supports two setup modes:
+
+- Full hosting: MAUI controls are rendered through Avalonia (`UseAvaloniaApp`).
+- Embedding: only `AvaloniaView` is registered inside a MAUI-native app (`UseAvaloniaEmbedding<TApp>`).
+
+Before your app can use Avalonia-drawn controls, fonts, or assets, register the appropriate services in `MauiProgram.cs` by chaining extension methods on `MauiAppBuilder` inside `CreateMauiApp`.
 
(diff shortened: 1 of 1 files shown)


docs/config-and-setup.md
```diff
@@ -0,0 +1,319 @@
...
+The equivalent C# code is:
+
+```csharp
+// Use font alias
+var label = new Label
+{
+    Text = "Hello",
+    FontFamily = "OpenSansRegular"
+};
+```
```
drasticactions, line 246: The Maui Asset config is the same as it is in MAUI itself, so IMO I would point to its documentation for how it works, and mention how we integrate into it through AvaloniaResource and the Essentials APIs.

docs/config-and-setup.md
```diff
@@ -0,0 +1,319 @@
...
+    {
+        // Update handler.PlatformView to reflect view.Value
+    }
+
+    private static void MapMaximum(RatingViewHandler handler, RatingView view)
+    {
+        // Update handler.PlatformView to reflect view.Maximum
+    }
+}
+```
```
drasticactions, line 348: This could be its own document, and it needs to be made clear that it needs either `Avalonia.Controls.Maui.Handlers.ViewHandler` for Avalonia based components, or `Avalonia.Controls.Maui.Handlers.AvaloniaControlHandler` ones that can be Maui native or Avalonia (This also needs an API review since I'm not happy with these names and it would be nice to have better ones)

It could be a full example showing an end to end process for how to do it. This wouldn't be enough. And I believe the root interfaces our methods inherent from use Internals in MAUI that a normal user can't access, which we need to iterate on for future versions.

I would drop this section for now.

drasticactions:
@jsuarezruiz It's fine here. I'm going to add docs to the [docs](https://github.com/AvaloniaUI/avalonia-docs) repo but having md's here is fine too. I assume an LLM would be able to pick those up and get a high level reading of the repo with it. Might even be worth doing the `SKILLS` thing I've seen around too at some point.

drasticactions (approved):

PR #156, opened 2026-03-09
Implement Label TextType property

jsuarezruiz:
Implement Label **TextType** Property https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.label.texttype.  MAUI uses `NSAttributedString` (iOS) or `Html.fromHtml `(Android) which support roughly the same subset. The main tags MAUI officially documents for TextType.Html are `<b>`, `<i>`, `<u>`, `<br>`, `<p>`, `<a>`, `<font>`, and we go beyond that.

image


drasticactions (commented):
> TextType.Html are `<b>, <i>, <u>, <br>, <p>, <a>, <font>, and we go beyond that.`

I need to think about this more, but if we're going to support this, then I think it should only support what MAUI Native can support. If we have additional support that MAUI itself can't do, then someone using it here can't use it in MAUI native itself, and they may want to have Native versions for some platforms, and drawn for others. And if we go _too_ overboard (Like supporting buttons), people will try to use it for UI itself which I _really_ wouldn't want. In the past when I've known APIs to use HTML as output, it was for social media like Twitter for posts. That may still be the case today, but even in those cases most of those services use facets or byte slices for people to create native UI with, rather than HTML to be embedded.

I appreciate this, and I know it's a supported feature in MAUI, but IMO I feel like if you're someone who wants HTML labels for UI, you could vibe that yourself at this point and make it work for your app, rather than us provide an implementation that you can't customize to your needs.

I need to sleep on it and think about it some more.

jsuarezruiz:
> > TextType.Html are `<b>, <i>, <u>, <br>, <p>, <a>, <font>, and we go beyond that.`
> 
> I need to think about this more, but if we're going to support this, then I think it should only support what MAUI Native can support. If we have additional support that MAUI itself can't do, then someone using it here can't use it in MAUI native itself, and they may want to have Native versions for some platforms, and drawn for others. And if we go _too_ overboard (Like supporting buttons), people will try to use it for UI itself which I _really_ wouldn't want. In the past when I've known APIs to use HTML as output, it was for social media like Twitter for posts. That may still be the case today, but even in those cases most of those services use facets or byte slices for people to create native UI with, rather than HTML to be embedded.
> 
> I appreciate this, and I know it's a supported feature in MAUI, but IMO I feel like if you're someone who wants HTML labels for UI, you could vibe that yourself at this point and make it work for your app, rather than us provide an implementation that you can't customize to your needs.
> 
> I need to sleep on it and think about it some more.



> > TextType.Html are `<b>, <i>, <u>, <br>, <p>, <a>, <font>, and we go beyond that.`
> 
> I need to think about this more, but if we're going to support this, then I think it should only support what MAUI Native can support. If we have additional support that MAUI itself can't do, then someone using it here can't use it in MAUI native itself, and they may want to have Native versions for some platforms, and drawn for others. And if we go _too_ overboard (Like supporting buttons), people will try to use it for UI itself which I _really_ wouldn't want. In the past when I've known APIs to use HTML as output, it was for social media like Twitter for posts. That may still be the case today, but even in those cases most of those services use facets or byte slices for people to create native UI with, rather than HTML to be embedded.
> 
> I appreciate this, and I know it's a supported feature in MAUI, but IMO I feel like if you're someone who wants HTML labels for UI, you could vibe that yourself at this point and make it work for your app, rather than us provide an implementation that you can't customize to your needs.
> 
> I need to sleep on it and think about it some more.



> > TextType.Html are `<b>, <i>, <u>, <br>, <p>, <a>, <font>, and we go beyond that.`
> 
> I need to think about this more, but if we're going to support this, then I think it should only support what MAUI Native can support. If we have additional support that MAUI itself can't do, then someone using it here can't use it in MAUI native itself, and they may want to have Native versions for some platforms, and drawn for others. And if we go _too_ overboard (Like supporting buttons), people will try to use it for UI itself which I _really_ wouldn't want. In the past when I've known APIs to use HTML as output, it was for social media like Twitter for posts. That may still be the case today, but even in those cases most of those services use facets or byte slices for people to create native UI with, rather than HTML to be embedded.
> 
> I appreciate this, and I know it's a supported feature in MAUI, but IMO I feel like if you're someone who wants HTML labels for UI, you could vibe that yourself at this point and make it work for your app, rather than us provide an implementation that you can't customize to your needs.
> 
> I need to sleep on it and think about it some more.

Mnn, good point around not allow additional support and just allow what MAUI itself can do. Will refactor, so, next time when you check will be the same as MAUI.

jsuarezruiz:
Aligned with MAUI.  Removed from converter: `<h1-h6>`, `<div>/<blockquote>`, `<code>/<tt>`.                                                                                                                                Kept (MAUI parity): `<b>/<strong>`, `<i>/<em>`, `<u>`, `<br>`, `<p>`, `<a>`, `<font>` (color + size)
Also, added a render test.

PR #171, opened 2026-03-23
Implement Clipboard Essentials API

jsuarezruiz:
Changes:
 - **Clipboard** (`IClipboard`): Full implementation with platform-specific paths for desktop (Avalonia TopLevel.Clipboard + Window.Activated change detection) and browser (navigator.clipboard via [JSImport])

Tested:
image


src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            await PlatformSetTextAsync(text);
+        }
+        catch
+        {
+            return;
+        }
+
+        UpdateState(text);
+        ClipboardContentChanged?.Invoke(this, EventArgs.Empty);
```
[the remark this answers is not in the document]
jsuarezruiz: Fixed. `PlatformSetTextAsync` now returns bool to signal success. `UpdateState` and `ClipboardContentChanged` only execute when the platform call actually modified the clipboard. When no TopLevel/clipboard is available, the desktop implementation returns false and the event is not raised.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Default.cs
```diff
@@ -0,0 +1,80 @@
...
+            return await Dispatcher.UIThread.InvokeAsync(async () => await clipboard.TryGetTextAsync());
+
+        return await clipboard.TryGetTextAsync();
+    }
+
+    private partial async Task PlatformSetTextAsync(string? text)
+    {
+        var clipboard = GetClipboard();
+        if (clipboard is null)
+            return;
```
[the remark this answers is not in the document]
jsuarezruiz: Fixed.

src/Avalonia.Controls.Maui.Essentials/MainThread/AvaloniaMainThread.cs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// <summary>
+    /// Invokes an action on the main UI thread. If already on the main thread, the action is executed immediately.
+    /// </summary>
+    /// <param name="action">The action to invoke on the main thread.</param>
+    public static void BeginInvokeOnMainThread(Action action)
+    {
+        if (IsMainThread)
+            action();
+        else
```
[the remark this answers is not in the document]
jsuarezruiz: Fixed.

src/Avalonia.Controls.Maui.targets
```diff
@@ -160,4 +160,18 @@
...
+          Condition="'$(AvaloniaControlsMauiDisableMainThreadAlias)' != 'true'
+              And ('$(_AvMauiTargetPlatformId)' == '' Or '$(_AvMauiTargetPlatformId)' == 'browser')">
+    <ItemGroup>
+      <_EssentialsRef Include="@(ProjectReference)" Condition="$([System.String]::new('%(Filename)').Contains('Avalonia.Controls.Maui.Essentials'))" />
+      <_EssentialsRef Include="@(PackageReference)" Condition="$([System.String]::new('%(Identity)').Contains('Avalonia.Controls.Maui.Essentials'))" />
+    </ItemGroup>
+    <ItemGroup Condition="'@(_EssentialsRef)' != ''">
+      <Using Include="Avalonia.Controls.Maui.Essentials.AvaloniaMainThread" Alias="MainThread" />
+    </ItemGroup>
+  </Target>
```
drasticactions: This is very clever, but I don't think it's worth it. We should go straight to fixing this in .NET MAUI and adding a proper way to set the MainThread implementation.

drasticactions (changes requested):
We shouldn't introduce a hack for MainThread, but Clipboard is fine. If you remove the MainThread code, then I'm 👍 .

src/Avalonia.Controls.Maui.targets
jsuarezruiz: Agree, that's the best approach. Question, should we add a TODO here with a link to an issue in the .NET MAUI repo and remove it once fixed?

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Browser.cs
```diff
@@ -0,0 +1,22 @@
+namespace Avalonia.Controls.Maui.Essentials;
+
+/// <summary>
+/// Browser-specific clipboard implementation using direct <c>navigator.clipboard</c> JavaScript interop
+/// for reliable clipboard access on WebAssembly targets.
+/// </summary>
+public partial class AvaloniaClipboard
+{
+    private partial Task<string?> PlatformGetTextAsync()
```
maxkatz6: Is there a reason why we don't use Avalonia APIs there?

maxkatz6: I.e. if yes, it should be documented in code

drasticactions: Ah, I didn't notice the split here. The default APIs in Avalonia work fine in WASM for this; there's no reason to split that.

In general, with Essentials APIs, it should always be to use Avalonia methods first. If it's a concept that doesn't exist in Avalonia, then we should think about adding it there first as a thing before doing anything custom here.

src/Avalonia.Controls.Maui.targets
jsuarezruiz: Ok, removed logic around `MainThread`.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Browser.cs
jsuarezruiz: Use Avalonia APIs here directly.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        if (_isSubscribed)
+            return;
+
+        if (_platformProvider.GetTopLevel() is Window window)
+        {
+            window.Activated += OnWindowActivated;
+            _isSubscribed = true;
+        }
+    }
```
drasticactions, line 52: I'm not sure if this is correct. `EnsureSubscribedToWindowActivation` runs whenever `GetClipboard` is called, which occurs when a user explicitly calls a clipboard API. If they had an event handle for ClipboardContentChanged, it will only ever run if they happen to call one of these methods first.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    void EnsureSubscribedToWindowActivation()
+    {
+        if (_isSubscribed)
+            return;
+
+        if (_platformProvider.GetTopLevel() is Window window)
+        {
+            window.Activated += OnWindowActivated;
```
drasticactions, line 48: `window.Activated` is attached to, but nothing disposes of it.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
```diff
@@ -0,0 +1,78 @@
...
+            var clipboard = _platformProvider.GetTopLevel()?.Clipboard;
+            if (clipboard is null)
+                return;
+
+            var currentText = await clipboard.TryGetTextAsync();
+            var changed = currentText != _lastKnownText;
+            UpdateState(currentText);
+
+            if (changed)
+                ClipboardContentChanged?.Invoke(this, EventArgs.Empty);
```
drasticactions: If you're in the app and you copy text from another text field, without activating the window, this event won't get invoked. It only activates when the window is activated.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
jsuarezruiz: Fixed. Applied changes and storing the subscribed Window reference and unsubscribing both Activated and Closed handlers in OnWindowClosed.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
jsuarezruiz: `EnsureSubscribed` now runs from the `ClipboardContentChanged` add accessor instead of `GetClipboard`.

Directory.Build.props
```diff
@@ -54,6 +54,11 @@
         <MicrosoftExtensionsVersion>11.0.0-preview.3.26207.106</MicrosoftExtensionsVersion>
     </PropertyGroup>
 
+    <ItemGroup Condition="'$(MSBuildProjectName)' == 'Avalonia.Controls.Maui.Desktop'">
+        <!-- Avalonia.FreeDesktop still declares the vulnerable version; force the patched one. -->
+        <PackageReference Include="Tmds.DBus.Protocol" Version="0.92.0" />
+    </ItemGroup>
+
```
drasticactions, line 61: This shouldn't be needed with the version bumps made elsewhere, and if something like this happens in the future, we should always fix the root versions rather than work around them.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// Initializes a new instance of the <see cref="AvaloniaClipboard"/> class with the specified platform provider.
+    /// </summary>
+    /// <param name="platformProvider">The platform provider used to retrieve the Avalonia TopLevel for clipboard access.</param>
+    public AvaloniaClipboard(IAvaloniaEssentialsPlatformProvider platformProvider)
+    {
+        _platformProvider = platformProvider;
+    }
+
+    /// <inheritdoc/>
+    public bool HasText => _hasText;
```
drasticactions, line 29: Checking in the maui repo, these values are computed when called via OS values, (Ex, https://github.com/dotnet/maui/blob/388b82f166d937c69a4553556b0042f441eadcd6/src/Essentials/src/Clipboard/Clipboard.ios.cs#L20-L21)

Here, it's cached, so as far as I can tell, SetTextAsync, GetTextAsync, or Activated would need to be called to update the value. Since Activate _should_ be called when you touch the window, that _might_ be fine, but there might be a better way to handle this. @maxkatz6 @MrJul, what do you think about this?

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
```diff
@@ -0,0 +1,88 @@
...
+    private partial async Task<string?> PlatformGetTextAsync()
+    {
+        var clipboard = _platformProvider.GetTopLevel()?.Clipboard;
+        if (clipboard is null)
+            return null;
+
+        if (!Dispatcher.UIThread.CheckAccess())
+            return await Dispatcher.UIThread.InvokeAsync(async () => await clipboard.TryGetTextAsync());
+
+        return await clipboard.TryGetTextAsync();
```
maxkatz6, line 24: It was a late addition to the Avalonia 12, but where possible, we should use control specific Dispatcher.
So, instead of global `UIThread`, it should be `topLevel.Dispatcher`.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.cs
maxkatz6: With async-only nature of current clipboard implementation, it would be hard to implement something like this.
On desktop, you we can use DispatcherFrame hack to (probably) safely run async code in-sync. 

Otherwise, always returning `true` might be the safest bet (there is never a guarantee clipboard is not cleared between HasText and GeText calls anyway).

We need to have a new API on the Avalonia side, for clipboard changed events.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
maxkatz6: But that can be fixed in another PR, as I assume there are more places like this.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.cs
drasticactions: > We need to have a new API on the Avalonia side, for clipboard changed events.

Yeah, that sounds good. It may be best to hold off on this PR for now and implement it in Avalonia first, then here. I don't think this is so important that we need to hack around it right now for the sake of the previews. We have time to implement it properly in Avalonia and use that instead.

src/Avalonia.Controls.Maui.Essentials/Clipboard/AvaloniaClipboard.Platform.cs
drasticactions: Made an issue for this, https://github.com/AvaloniaUI/Avalonia.Controls.Maui/issues/188

It's in a bunch of places, and we can refactor it.

PR #173, opened 2026-03-25
Refactor Memory Benchmarks

jsuarezruiz:
I started on this after one (`MenuAndToolbarLeakBenchmark`) of the memory leak benchmarks looked flaky.

This OR apply changes to use `NativeMemoryEstimate` (PrivateMemoryDelta - GcCommittedBytesDelta), so the failure signal is closer to what we actually care about.

It also centralizes the threshold logic in BenchmarkTestPage and slightly relaxes the GC wait pattern in `MenuAndToolbarLeakBenchmark` to reduce timing noise on slower runners.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,6 +35,7 @@ jobs:
               - 'build/**'
             test:
               - 'tests/Avalonia.Controls.Maui.*/**'
+              - 'samples/BenchmarkApp/**'
               - 'Directory.*.props'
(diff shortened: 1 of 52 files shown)


drasticactions (approved):
LGTM. I disabled the WebView test for now since I need to validate that locally vs. CI, since I can't replicate that locally on my Arch setup and I don't know how that works on Ubuntu (where I'm 99% sure it doesn't have WPE setup)

PR #189, opened 2026-04-27
Fix duplicate CollectionView selection issue

jsuarezruiz:
This PR fixes a duplicate navigation bug caused by `CollectionViewHandler.OnSelectionChanged` manually executing `SelectionChangedCommand` after synchronizing selection state, which resulted in the command firing twice in scenarios like Shell gallery item selection because MAUI already invokes the command when SelectedItem/SelectedItems changes; the fix removes the handler-level manual command execution and keeps only the state synchronization logic for single and multiple selection modes, and the related handler tests were tightened to assert exactly one command invocation so this regression is caught going forward.

Now, in the MAUI Gallery, not have double navigation issue:
fix-navigation-gallery

### DIFF
--- a/src/Avalonia.Controls.Maui/Handlers/CollectionViewHandler.cs
+++ b/src/Avalonia.Controls.Maui/Handlers/CollectionViewHandler.cs
@@ -153,15 +153,11 @@ public class CollectionViewHandler : ViewHandler<CollectionView, MauiCollectionV
                 }
                 else if (currentSelectableItemsView.SelectionMode == Microsoft.Maui.Controls.SelectionMode.Multiple)
                 {
-                    if (!Equals(currentSelectableItemsView.SelectedItem, selectedItem))
-                    {
-                        currentSelectableItemsView.SelectedItem = selectedItem;
-                    }
-
+                    var desiredSelectedItems = selectedItems ?? [];
                     var virtualSelectedItems = currentSelectableItemsView.SelectedItems;
-                    if (virtualSelectedItems != null)
+                    if (!SelectionEquals(virtualSelectedItems, desiredSelectedItems))
                     {
-                        SynchronizeSelectedItems(virtualSelectedItems, selectedItems ?? []);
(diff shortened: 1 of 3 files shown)


src/Avalonia.Controls.Maui/Handlers/CollectionViewHandler.cs
```diff
@@ -131,43 +131,37 @@ private void OnSelectionChanged(object? sender, EventArgs e)
...
                     }
 
-                    // Execute command for Multiple selection (pass the list of items)
-                    var parameter = selectableItemsView.SelectionChangedCommandParameter ?? virtualSelectedItems;
-                    if (selectableItemsView.SelectionChangedCommand?.CanExecute(parameter) == true)
+                    var virtualSelectedItems = currentSelectableItemsView.SelectedItems;
+                    if (virtualSelectedItems != null)
                     {
-                        selectableItemsView.SelectionChangedCommand.Execute(parameter);
+                        SynchronizeSelectedItems(virtualSelectedItems, selectedItems ?? []);
```
drasticactions: I think this breaks `SelectionMode.Multiple`. If you enable it and select multiple items, it will only select the last item selected. Open the ControlGallery and run the Multiple Selection example in the CollectionViewPage.

drasticactions (changes requested):

src/Avalonia.Controls.Maui/Handlers/CollectionViewHandler.cs
jsuarezruiz: Good catch. Multiple selection was working but syncing through SelectedItem caused MAUI to raise a one-item `SelectionChanged` event. I fixed that by switching the handler to MAUI’s multiple-selection path and updating the virtual view with UpdateSelectedItems().

I also added a test that drives the actual platform selection path and verifies that `SelectionChangedEventArgs.CurrentSelection` preserves the full multi-selection set. 

cv-multi-selection

drasticactions (approved):

PR #199, opened 2026-05-05
Added Essentials Geolocation

cedric56:
This PR adds a complete Geolocation implementation for Avalonia applications using the Essentials pattern, with cross-platform support for Linux and WebAssembly.

🐧 Linux: Leverages the powerful Geoclue service, a standard location framework on the Linux desktop. This ensures your apps can access location data via GPS, Wi-Fi, or other sources provided by the system.

🕸️ WebAssembly (Wasm): Uses the browser's built-in Geolocation API. This means your Avalonia web apps can now request and receive location directly from the user's browser, just like a standard web application.


jsuarezruiz:
Thanks for the contribution and for pushing Essentials geolocation forward. The direction makes sense: GeoClue is the right Linux backend and the browser Geolocation API is the natural WASM backend.

src/Avalonia.Controls.Maui.Essentials/Geolocation/AvaloniaGeolocation.cs
```diff
@@ -0,0 +1,60 @@
...
+        PlatformStopListeningForeground();
+    }
+
+    internal void OnLocationChanged(MauiLocation location) =>
+            OnLocationChanged(new GeolocationLocationChangedEventArgs(location));
+
+    internal void OnLocationChanged(GeolocationLocationChangedEventArgs e) =>
+        LocationChanged?.Invoke(null, e);
+
+    internal void OnLocationError(GeolocationError geolocationError) =>
```
jsuarezruiz: MAUI documents `ListeningFailed` as terminal for the current listening session, but this only raises the event. Browser errors and D-Bus callback failures can leave the watch/subscription active, so `IsListeningForeground` remains true and later calls can fail with `Already listening`. Could the platform listener be stopped/reset before raising `ListeningFailed`?

src/Avalonia.Controls.Maui.Essentials/Avalonia.Controls.Maui.Essentials.csproj
```diff
@@ -23,6 +23,13 @@
     <PackageReference Include="Microsoft.Maui.Core" Version="$(MauiVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'browser'">
+    <PackageReference Include="Tmds.DBus.Protocol" Version="0.92.0" />
```
jsuarezruiz, line 27: The current package adds `Tmds.DBus.Protocol` to the shared `net11.0` dependency group. That asset can be used by Windows and macOS consumers too, even though only Linux uses GeoClue.

This probably needs a packaging/design decision rather than just `IsOSPlatform('Linux')`, because that condition is evaluated when building/packing the package. Could we avoid adding the D-Bus dependency to non-Linux consumers, or explicitly document that the desktop asset carries this Linux-only dependency?

samples/ControlGallery/ControlGallery/Properties/launchSettings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "profiles": {
+    "ControlGallery": {
+      "commandName": "Project"
+    },
+    "WSL": {
+      "commandName": "WSL2",
+      "distributionName": ""
```
jsuarezruiz: This PR adds a sample `launchSettings.json` with a WSL profile and an empty `distributionName`. That seems developer-local and unrelated to geolocation.

Could this be removed from the PR unless it is required for the sample to run?

src/Avalonia.Controls.Maui.Essentials/Avalonia.Controls.Maui.Essentials.csproj
maxkatz6: It would be problematic, as "net11.0-linux" target doesn't exist. So, adding it to a generic target ("net11.0") is fine, assuming code is trimmable and only used behind OperatingSystem.IsLinux().

src/Avalonia.Controls.Maui.Essentials/Geolocation/AvaloniaGeolocation.cs
cedric56: I add a call to PlatformStopListeningForeground before raising event.

samples/ControlGallery/ControlGallery/Properties/launchSettings.json
cedric56: File removed

src/Avalonia.Controls.Maui.Essentials/Avalonia.Controls.Maui.Essentials.csproj
cedric56: So i leave it as it is ?

jsuarezruiz (commented):
Could you rebase and fix the conflict? Thanks in advance.

PR #200, opened 2026-05-07
Added Essentials AppActions

cedric56:
✨ Add AppActions support for Avalonia MAUI Essentials (Linux D-Bus integration)

This PR introduces a new AppActions implementation for Avalonia.Controls.Maui.Essentials, enabling Linux desktop integration through D-Bus activation and .desktop file actions.

🚀 Overview

This feature extends MAUI AppActions to support desktop Linux environments when using Avalonia, allowing applications to expose and handle native app shortcuts via:

D-Bus service registration
Dynamic .desktop file generation
Runtime action activation callbacks

✅ Linux Platform Support

Implements PlatformSetAsync for Linux only
Uses OperatingSystem.IsLinux() guard

🔌 D-Bus Integration

Registers a session bus service (com.essentials.AppActions)
Exposes actions via Tmds.DBus
Handles activation through OnAppActionAsync

🖥 Desktop Integration

Generates .desktop files dynamically per app instance
Adds custom Desktop Action entries for each AppAction
Uses gdbus calls to trigger actions externally

🪟 Window Integration

Sets X11 WM_CLASS to support desktop identity binding
Ensures initialization occurs after MainWindow is available


src/Avalonia.Controls.Maui.Essentials/Avalonia.Controls.Maui.Essentials.csproj
```diff
@@ -23,6 +23,12 @@
     <PackageReference Include="Microsoft.Maui.Core" Version="$(MauiVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'browser'">
+    <PackageReference Include="Tmds.DBus" Version="0.92.0">
+      <Aliases>TmdsDBus</Aliases>
+    </PackageReference>
```
jsuarezruiz, line 29: This condition excludes only the browser TFM, so the packed `net11.0` dependency group now includes `Tmds.DBus` for every desktop consumer, including Windows and macOS, even though the implementation only supports Linux. This needs a packaging/design decision: avoid flowing the Linux-only dependency to non-Linux consumers, or explicitly accept and document that the shared desktop asset carries it.

samples/ControlGallery/ControlGallery/Properties/launchSettings.json
```diff
@@ -0,0 +1,11 @@
...
+  "profiles": {
+    "ControlGallery": {
+      "commandName": "Project"
+    },
+    "WSL": {
+      "commandName": "WSL2",
+      "distributionName": ""
+    }
+  }
+}
```
jsuarezruiz: This adds a sample `launchSettings.json` with a WSL profile and an empty `distributionName`. It appears developer-local and unrelated to AppActions support, so it should be removed unless the PR requires it for a reproducible sample scenario.

src/Avalonia.Controls.Maui.Essentials/AppActions/AvaloniaAppActions.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            return Task.FromResult(_actions ??= Array.Empty<AppAction>());
+        }
+
+        public Task SetAsync(IEnumerable<AppAction> actions)
+        {
+            if (!IsSupported)
+                throw new FeatureNotSupportedException();
+
+            _actions = actions;
```
jsuarezruiz: `SetAsync` updates `_actions` before `PlatformSetAsync` completes. If the platform registration fails after this point, `GetAsync()` will report actions that were never actually installed, and callers cannot tell that the in-memory state no longer matches the desktop integration. Could `_actions` be assigned only after `PlatformSetAsync(actions)` succeeds?

cedric56: _actions is not assign after PlatformSetAsync succeeds.

samples/ControlGallery/ControlGallery/Properties/launchSettings.json
cedric56: File removed

src/Avalonia.Controls.Maui.Essentials/AppActions/AvaloniaAppActions.Default.cs
```diff
@@ -0,0 +1,262 @@
...
+            {
+                if (Application.Current?.ApplicationLifetime == null)
+                    throw new InvalidOperationException("ApplicationLifetime is not initialized");
+
+                if (Application.Current?.ApplicationLifetime is not IClassicDesktopStyleApplicationLifetime desktop)
+                    throw new InvalidOperationException("Only IClassicDesktopStyleApplicationLifetime is supported");
+
+                // We create a new instance of the Current class, to get a unique WMClass
+                // this will force refreshing shortcuts in the desktop environment
+                var current = new Current(desktop, actions, AppActionActivated);
```
jsuarezruiz, line 49: Every `SetAsync` creates a new D-Bus service/object/desktop file and only cleans up on app exit. Updating/removing actions can leave stale actions?

src/Avalonia.Controls.Maui.Essentials/AppActions/AvaloniaAppActions.Default.cs
```diff
@@ -0,0 +1,262 @@
...
+
+                // Base desktop entry
+                string content = $@"[Desktop Entry]
+Name={name}
+Exec={dotnet} {dll} {uid}
+Type=Application
+Categories=Utility;
+StartupNotify=true
+StartupWMClass={uid}
+Actions={string.Join(";", actions.Select(a => a.Title))};
```
jsuarezruiz: Titles are user-facing text and can contain spaces, semicolons, or other characters that could break desktop-entry syntax. This could use sanitized stable action IDs for desktop action identifiers, and quote/escape action IDs in the generated `Exec` command.

jsuarezruiz (changes requested):
Could you rebase and fix the conflict? Thanks in advance.

jsuarezruiz:
CI is currently failing in the ControlGallery NativeAOT publish because `AppActionsPage.xaml` uses string-path bindings inside the `CollectionView` item template.

The generated XAML code calls `Microsoft.Maui.Controls.Binding.Binding(string, ...)`, which is marked `RequiresUnreferencedCode`, so NativeAOT/trimming fails with IL2026:

`Using bindings with string paths is not trim safe. Use expression-based binding instead.`

Can you review?

src/Avalonia.Controls.Maui.Essentials/AppActions/AvaloniaAppActions.Default.cs
cedric56: I added a class SanitizedClass to handle id

src/Avalonia.Controls.Maui.Essentials/AppActions/AvaloniaAppActions.Default.cs
cedric56: Actions can only be updated with SetAppActions method which associate a new guid to the window so all actions defined previously on a .desktop file are still but cannot reach the window.

cedric56:
> CI is currently failing in the ControlGallery NativeAOT publish because `AppActionsPage.xaml` uses string-path bindings inside the `CollectionView` item template.
> 
> The generated XAML code calls `Microsoft.Maui.Controls.Binding.Binding(string, ...)`, which is marked `RequiresUnreferencedCode`, so NativeAOT/trimming fails with IL2026:
> 
> `Using bindings with string paths is not trim safe. Use expression-based binding instead.`
> 
> Can you review?

I added a x:DataType to CollectionView ItemTemplate.

cedric56 (commented):
I have added x:DataType="model:AppAction"

PR #205, opened 2026-05-12
Applied AvaloniaFileResult to MediaPicker

cedric56:
Finishing for MediaPicker as well https://github.com/AvaloniaUI/Avalonia.Controls.Maui/issues/202


drasticactions:
Thank you for catching my mistake! I didn't notice it also used FileResult, and I agree it would make sense to use it for both. Since .NET 11 Preview 4 is out, and I should have implemented it myself in the issue you brought up, I did this myself in https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/206, so I'm going to close this one.

PR #210, opened 2026-05-21
Implement EntryCell VerticalTextAlignment

jsuarezruiz:
Adds `EntryCell.VerticalTextAlignment` support in the compatibility handler.

Changes:
- Maps `EntryCell.VerticalTextAlignment` through the existing `EntryCellHandler` mapper.
- Adds test coverage for initial mapping and property updates. Also, added a render test.
- Updates the sample to show `EntryCell` vertical alignment behavior.


drasticactions:
Note for me in the future: Avalonia doesn't support Vertical Text Alignment, so this is a hack that uses Vertical Content Alignment. The PR is consolidating an existing hack that implements it for other things. Vertical Text Alignment should really be a feature in Avalonia if we want to support it here, since this will end up biting us at some point.

drasticactions (approved):

PR #214, opened 2026-06-14
Fix InvalidCastException when wrapping TabbedPage/NavigationPage in browser target

ibrahimelkady2001:
Fixes two `InvalidCastException` crashes when using `TabbedPage` with `NavigationPage` children on the browser (Avalonia) target.

### 1. `TabbedPageExtensions.UpdateChildren` — `NavigationPage` children crash
When `TabbedPage` contains `NavigationPage` children (e.g. `Children.Add(new NavigationPage(new HomePage()) { Title = "Home" })`), the `UpdateChildren` method at line 118 did:
```csharp
var wrappedPage = (AvaloniaContentPage)page.ToPlatform(mauiContext);
```
`NavigationPage.ToPlatform()` returns `AvaloniaNavigationPage` — which is NOT `AvaloniaContentPage` — causing `InvalidCastException`.

**First attempted fix:** Unwrap `NavigationPage` to `RootPage` before calling `ToPlatform()`. This fixed the crash but **broke per-tab navigation** (`Navigation.PushAsync()` stopped working because each tab was just a plain `ContentPage` without a navigation stack).

**Final fix:** Keep the `NavigationPage` wrapper intact. `AvaloniaNavigationPage` extends `Avalonia.Controls.Page`, so use that as the base type instead of forcing `AvaloniaContentPage`:

```csharp
if (page is Microsoft.Maui.Controls.NavigationPage)
{
    wrappedPage = (AvaloniaPage)page.ToPlatform(mauiContext);
}
else
{
    // Normal ContentPage — safe to cast to AvaloniaContentPage
    ...
}
```

This preserves the NavigationPage stack so per-tab push/pop navigation continues to work.

### 2. `StackNavigationManager.WrapPage` — TabbedPage/FlyoutPage crash
When navigating to a `TabbedPage` or `FlyoutPage`, `WrapPage` did:
```csharp
return (AvaloniaContentPage)mauiPage.ToPlatform(MauiContext);
```
These page types return `AvaloniaTabbedPage`/`AvaloniaFlyoutPage` which aren't `AvaloniaContentPage`. Fix: use safe cast with fallback wrapping:

```csharp
var platformPage = mauiPage.ToPlatform(MauiContext);
if (platformPage is AvaloniaContentPage cp) return cp;
if (platformPage is Avalonia.Controls.Control control)
    return new AvaloniaContentPage { Content = control };
```

### Changes:

**`StackNavigationManager.cs`** — `WrapPage` now safely casts non-ContentPage types

**`TabbedPageExtensions.cs`** — `UpdateChildren` now:
- Checks if child is `NavigationPage` and keeps it as-is (preserving navigation stack)
- Uses `AvaloniaPage` base class instead of `AvaloniaContentPage` for non-ContentPage children
- `CreateTabHeader(page)` still uses the original page (NavigationPage) for title/icon

### Stack traces fixed:

```
System.InvalidCastException: Arg_InvalidCastException
   at Avalonia.Controls.Maui.Platform.StackNavigationManager.WrapPage(IView mauiPage)
   at Avalonia.Controls.Maui.Platform.StackNavigationManager.NavigateTo(NavigationRequest request)
   at Avalonia.Controls.Maui.Handlers.NavigationViewHandler.RequestNavigation(...)
```

```
System.InvalidCastException: Arg_InvalidCastException
   at Avalonia.Controls.Maui.Extensions.TabbedPageExtensions.UpdateChildren(...)
   at Avalonia.Controls.Maui.Handlers.TabbedPageHandler.MapItemsSource(...)
```

### DIFF
--- a/src/Avalonia.Controls.Maui/Platform/StackNavigationManager.cs
+++ b/src/Avalonia.Controls.Maui/Platform/StackNavigationManager.cs
@@ -612,7 +612,12 @@ public class StackNavigationManager
     /// <returns>The Avalonia ContentPage.</returns>
     protected virtual AvaloniaContentPage WrapPage(IView mauiPage)
     {
-        return (AvaloniaContentPage)mauiPage.ToPlatform(MauiContext);
+        var platformPage = mauiPage.ToPlatform(MauiContext);
+        if (platformPage is AvaloniaContentPage cp)
+            return cp;
+        if (platformPage is Avalonia.Controls.Control control)
+            return new AvaloniaContentPage { Content = control };
+        throw new InvalidOperationException($"Cannot wrap page of type {platformPage.GetType()}");
     }
 
     /// <summary>
--- a/src/Avalonia.Controls.Maui/Extensions/TabbedPageExtensions.cs
+++ b/src/Avalonia.Controls.Maui/Extensions/TabbedPageExtensions.cs
@@ -115,7 +115,16 @@ public static class TabbedPageExtensions
 
         foreach (var page in mauiTabbedPage.Children)
         {
-            var wrappedPage = (AvaloniaContentPage)page.ToPlatform(mauiContext);
+            var contentPage = page is Microsoft.Maui.Controls.NavigationPage np ? np.RootPage : page;
+            var platformPage = contentPage.ToPlatform(mauiContext);
+            var wrappedPage = platformPage as AvaloniaContentPage;
+            if (wrappedPage == null)
+            {
+                if (platformPage is AvaloniaPage avPage)
+                    wrappedPage = new AvaloniaContentPage { Content = (Avalonia.Controls.Control)avPage };
+                else
+                    continue;
+            }
             wrappedPage.Header = CreateTabHeader(page);
             pages.Add(wrappedPage);
             pagesToLoadIcons.Add((wrappedPage, page));


src/Avalonia.Controls.Maui/Platform/StackNavigationManager.cs
```diff
@@ -612,7 +612,12 @@ private void FireNavigationFinished()
     /// <returns>The Avalonia ContentPage.</returns>
     protected virtual AvaloniaContentPage WrapPage(IView mauiPage)
     {
-        return (AvaloniaContentPage)mauiPage.ToPlatform(MauiContext);
+        var platformPage = mauiPage.ToPlatform(MauiContext);
+        if (platformPage is AvaloniaContentPage cp)
+            return cp;
+        if (platformPage is Avalonia.Controls.Control control)
+            return new AvaloniaContentPage { Content = control };
```
jsuarezruiz, line 619: `WrapPage` is called repeatedly for the same MAUI page, but this creates a new wrapper each time. That breaks the reference-based stack checks and can try to parent the same platform control twice. Can you review?

jsuarezruiz (commented):
This seems related with #221, could you verify? @drasticactions Thoughts?

jsuarezruiz:
@ibrahimelkady2001 Could you check the comments and also rebase to fix the conflicts? Thanks in advance!

PR #217, opened 2026-06-15
Fix OnStart lifecycle not firing on MAUI Application in Avalonia bridge

### LINKED ISSUE #216: Application.OnStart() lifecycle not firing on Avalonia browser target
## Bug Description

`MauiAvaloniaApplication.OnFrameworkInitializationCompleted()` sets up the MAUI `Application` handler via `SetApplicationHandler()` but **never calls `SendStart()`** on the MAUI `Application` instance.

In the standard MAUI platform (Android/iOS), the platform-specific application calls `Application.SendStart()` after initialization, which triggers the virtual `Application.OnStart()` method. The Avalonia bridge's `MauiAvaloniaApplication` is the equivalent class but is missing this call.

## Impact

Any MAUI `Application.OnStart()` overrides in user code never fire on the Avalonia browser target. This includes common startup logic like:
- Checking for saved authentication tokens
- Initializing app-wide services
- Restoring previous application state
- Any code placed in `App.OnStart()`

## Steps to Reproduce

1. Create a MAUI app with the Avalonia bridge for browser target
2. Override `OnStart()` in `App.xaml.cs`
3. Run on browser — `OnStart()` is never called

## Environment

- Avalonia.Controls.Maui version: 11.0.0-preview.4.26133.721
- Target: net11.0-browser

## Proposed Fix

PR #216 adds a call to `SendStart()` after `SetApplicationHandler()` in `MauiAvaloniaApplication.OnFrameworkInitializationCompleted()`:

```csharp
this.SetApplicationHandler(Application, ApplicationContext);
if (this.Application is Microsoft.Maui.Controls.Application startApp)
    startApp.SendStart();
```

ibrahimelkady2001:
## Problem

`MauiAvaloniaApplication.OnFrameworkInitializationCompleted()` sets up the MAUI `Application` handler via `SetApplicationHandler()` but never calls `SendStart()` on the MAUI `Application` instance. This means `Application.OnStart()` overrides in user code never fire on any Avalonia target (browser, desktop).

In the standard MAUI platform (Android/iOS), the platform-specific application calls `Application.SendStart()` after initialization. The Avalonia bridge's `MauiAvaloniaApplication` is the equivalent class but is missing this call.

## Fix

Add a call to `SendStart()` after `SetApplicationHandler()` in `MauiAvaloniaApplication.OnFrameworkInitializationCompleted()`:

```csharp
this.SetApplicationHandler(Application, ApplicationContext);
if (this.Application is Microsoft.Maui.Controls.Application startApp)
    startApp.SendStart();
```

## Impact

Fixes `Application.OnStart()` lifecycle firing on all Avalonia targets including browser (WASM), enabling common startup patterns like token restore, service initialization, and state restoration.

Closes #216

### DIFF
--- a/src/Avalonia.Controls.Maui/Platform/MauiAvaloniaApplication.cs
+++ b/src/Avalonia.Controls.Maui/Platform/MauiAvaloniaApplication.cs
@@ -106,6 +106,10 @@ public abstract class MauiAvaloniaApplication : Application, IPlatformApplicatio
         // Connect the MAUI Application to its handler
         this.SetApplicationHandler(Application, ApplicationContext);
 
+        // Fire OnStart lifecycle — the Avalonia bridge does not call it automatically
+        if (this.Application is Microsoft.Maui.Controls.Application startApp)
+            startApp.SendStart();
+
         if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
         {
             var avaloniaWindow = CreatePlatformWindow();
--- a/src/Avalonia.Controls.Maui/Platform/MauiAvaloniaApplication.cs
+++ b/src/Avalonia.Controls.Maui/Platform/MauiAvaloniaApplication.cs
@@ -106,10 +106,6 @@ public abstract class MauiAvaloniaApplication : Application, IPlatformApplicatio
         // Connect the MAUI Application to its handler
         this.SetApplicationHandler(Application, ApplicationContext);
 
-        // Fire OnStart lifecycle — the Avalonia bridge does not call it automatically
-        if (this.Application is Microsoft.Maui.Controls.Application startApp)
-            startApp.SendStart();
-
         if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
         {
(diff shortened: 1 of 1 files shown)


ibrahimelkady2001:
**Addressed review feedback:** Moved `SendStart()` from before window creation to after the window/content is created by the lifetime handler. This matches standard MAUI lifecycle order where `OnStart()` fires after the application window is initialized, so code in `OnStart()` can safely access `Application.Current.Windows` and the initial page.

drasticactions:
This would not have fixed the issue. Application.OnStart (and the related APIs your bot didn't implement) are internal in MAUI. While we have a hack in place to let us use internals, the goal is to _not_ use it long-term and try to get rid of those uses with public replacements. 

In this case, the right solution IMO is implementing IWindow Lifecycle events, which do call those internal APIs (although they are protected virtual methods in MAUI, so I don't _think_ you can access them long-term. We'll need to think about it.

https://github.com/AvaloniaUI/Avalonia.Controls.Maui/pull/220 superseeds this.

PR #219, opened 2026-06-15
Fix gesture recognizers not reconnecting after navigation pop

### LINKED ISSUE #218: Gesture recognizers inside CollectionView DataTemplate stop working after navigation pop
## Bug Description

Gesture recognizers (TapGestureRecognizer, SwipeGestureRecognizer, etc.) inside CollectionView DataTemplate items stop working after navigation pop back to the page.

## Impact

Any CollectionView with gesture recognizers in its DataTemplate (e.g., tapping an item to navigate, swipe to delete) breaks after the user navigates to another page and pops back. The gesture recognizers never fire again on the recycled item containers.

## Root Cause

`GestureManager.SetupGestureManager()` at line 82-87 disconnects gesture handlers when `_didHaveWindow` is `true` and `_view.Window` is `null`:

```csharp
if (handler == null || (_didHaveWindow && _view.Window == null))
{
    DisconnectGestures();
    return; // ← permanently disconnects gestures
}
```

During navigation pop, CollectionView item containers are temporarily detached from the visual tree. `WindowChanged` fires with `null` window, `_didHaveWindow` is `true`, so `DisconnectGestures()` runs and `SetupGestureManager()` returns early. After the pop completes and items are re-attached, the gesture handlers remain permanently disconnected because `SetupGestureManager()` is never called again to re-subscribe.

## Steps to Reproduce

1. Create a MAUI CollectionView with a DataTemplate containing a TapGestureRecognizer
2. Tap an item to navigate to another page  
3. Pop back to the original page
4. Tap any CollectionView item — gesture does not fire

## Environment

- Avalonia.Controls.Maui version: 11.0.0-preview.4.26133.721
- Target: net11.0-browser (WASM)

## Proposed Fix

PR #218 adds a check in `GestureManager.SetupGestureManager()`: when the handler is valid but the window is temporarily null during navigation transition, do not disconnect gesture handlers. Reset `_didHaveWindow` so the gesture manager can fully reconnect on next setup.

ibrahimelkady2001:
## Problem

`GestureManager.SetupGestureManager()` disconnects gesture handlers when `_didHaveWindow` is `true` and `_view.Window` is `null`:

```csharp
if (handler == null || (_didHaveWindow && _view.Window == null))
{
    DisconnectGestures();
    return; // permanently disconnects gestures, never reconnects
}
```

During navigation pop, CollectionView item containers are temporarily detached from the visual tree. `WindowChanged` fires with null window, `_didHaveWindow` is true, so `DisconnectGestures()` runs and `SetupGestureManager()` returns early. After the pop completes and items are re-attached, the gesture handlers remain permanently disconnected.

**Impact:** All gesture recognizers inside CollectionView DataTemplate items (TapGestureRecognizer, SwipeGestureRecognizer, etc.) stop working after navigate+pop. Page-level gesture recognizers are not affected because they go through a full page reconstruction cycle, while CollectionView recycles existing item containers.

## Fix

When the handler is valid but the window is null during a navigation transition, do not disconnect gesture handlers. Reset `_didHaveWindow` so the gesture manager can fully reconnect when the view is re-attached to the visual tree.

```csharp
if (handler != null && _view.Window == null)
{
    // Navigation transition — keep existing gesture handlers connected
    _didHaveWindow = false;
    return;
}
```

Closes #218

### DIFF
--- a/src/Avalonia.Controls.Maui/Platform/GestureManager.cs
+++ b/src/Avalonia.Controls.Maui/Platform/GestureManager.cs
@@ -82,6 +82,14 @@ internal class GestureManager : IDisposable
         if (handler == null ||
             (_didHaveWindow && _view.Window == null))
         {
+            // Navigation pop temporarily detaches views from the visual tree.
+            // Don't disconnect when handler is valid but window is null during transition —
+            // keep existing gesture handlers connected so they work when re-attached.
+            if (handler != null && _view.Window == null)
+            {
+                _didHaveWindow = false;
+                return;
+            }
             DisconnectGestures();
             return;
         }


drasticactions:
Gesture Recognizers are being refactored as part of https://github.com/dotnet/maui/pull/33365#issuecomment-4702963378 when it lands. Please have your bot wait until this work is done.

